PR #75, opened 2023-02-21
Implement caching for the Mastodon API

aumetra:


aumetra:
Proper invalidation will be implemented with #76

PR #77, opened 2023-02-21
Implement event system

### LINKED ISSUE #76: Introduce event emitters and consumers
## Proposal

An event-based architecture. Services that mutate the state, such as the post or account service, can emit events to consumers. These events notify about additions, updates, and deletions (and potentially more. This architecture should be kept extensible).  

This architecture is useful for:

- Cache invalidation
- WebSocket updates pushed to clients

## Implementation

The proposal would be implemented with Tokio broadcast channels in the beginning (since we are very much a monolithic application).  
Instead of directly exposing the channels to the clients though, I would prefer to expose something like an event consumer struct that gives you a stream that one can poll for updates.

The idea behind this is that future changes to how events are exchanged would be trivial if the implementation details are as hidden as possible.  
For example, we could switch to an RabbitMQ or Fluvio-based messaging backend in the future without having to make large code changes. The internals of the emitters and consumers may change, but that's all.

aumetra:
Closes #76

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2362,6 +2362,8 @@ dependencies = [
  "derive_builder 0.12.0",
  "futures-util",
  "redis",
+ "serde",
+ "serde_json",
  "tokio",
  "tokio-stream",
  "tracing",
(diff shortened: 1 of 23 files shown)


aumetra:
Tracking some further work based on this PR in #79

PR #83, opened 2023-02-25
Make search results chronologically ordered

aumetra:
Currently the results are ordered by their Levenshtein distance, i.e. the closer the search query matches the text, the higher it ranks.  
I'm not sure how Mastodon handles it, whether they do chronological handling, too. 

We should closely imitate Mastodon in terms of search result ordering, otherwise we might break expectations some clients have.

It would be great if someone on an instance with full-text search could chime in an tell me how Mastodon orders the results (my instance doesn't have full-text search for posts enabled)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -208,6 +208,8 @@ dependencies = [
  "memchr",
  "pin-project-lite",
  "tokio",
+ "zstd",
+ "zstd-safe",
 ]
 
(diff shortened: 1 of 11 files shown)


aumetra:
The API docs make it look like they're supposed to be chronologically sorted.  
Adding merging this for now.

PR #84, opened 2023-02-27
Add user and post pages

aumetra:
Nothing incredible, just something nice enough to look at.


aumetra:
Public post view:

image

PR #87, opened 2023-02-28
Update post database schema

### LINKED ISSUE #85: Change post database schema
Instead of storing the reposts split from the post, we should probably store them in the same table and identify reposts by having the `reposted_post_id` set to a value (not a finalised name)

This makes aggregation of home timelines easier, and makes it easier to implement a potential quote repost feature

aumetra:
Closes #85


kitsune/src/mapping/activitypub/activity.rs
```diff
@@ -116,6 +115,8 @@ impl IntoActivity for posts::Model {
     type NegateOutput = Activity;
 
     async fn into_activity(self, state: &Zustand) -> Result<Self::Output> {
+        // TODO: Decide type by `reposted_post_id` field
```
aumetra, line 118: Fix the TODOs

PR #89, opened 2023-02-28
Add ability to import posts and accounts via the search

aumetra:
You can now paste URLs into the search bar of your Mastodon client, and Kitsune will attempt to fetch the post/account


aumetra:
The Mastodon search has a bunch of logic I want to share with the eventual GraphQL search function.  
This should be somehow extracted into the search service.

aumetra:
Actually, there's not that much that can be abstracted away. I'll do it when I add the GraphQL function

PR #92, opened 2023-03-02
Add media attachment support

### LINKED ISSUE #31: Add media proxy
Add a media proxy to the backend. It shouldn't be too difficult to implement since every attachment, remote or local, gets an own ID in our database.

The path could be something like `/media-proxy/[attachment ID]` and it just fetches and streams the media to the client.  

This prevents IPs getting leaked to the remote server and allows admins to do some reverse-proxy level caching to decrease the bandwidth usage on the remote server

### LINKED ISSUE #86: Add media uploads
Add APIs for uploading media attachments.  
These APIs would probably go through a storage abstraction, since we might want to use multiple storage backends (similar to `kitsune-messaging`).  

Backends I definitely want to have are:

- File system
- S3

aumetra:
This PR adds the ability to upload media attachments to your posts.  
It also adds the `kitsune-storage` crate which is an opinionated abstraction over all kinds of storage backends, operating on string slices as paths and streams as a mean of transmitting data.

This PR adds support for the following storage backends:

- File system
- S3

Closes #86, Closes #31


aumetra:
We can basically alias the same API endpoint for `v2` since we currently don't do any fancy processing of uploads.  
Because of this we can always return 200, making it not really different from `v1`

PR #124, opened 2023-03-24
Basic OIDC support

### LINKED ISSUE #123: Support overarching OAuth provider
Say I wanna bundle Kitsune together with some other app-servers as part of a omg.lol-like service.

For this type of hydra-service, a unified login is necessary. The best open source provider of that atm seems to be https://www.ory.sh — They also have an actively (auto-)updated [Rust client](https://github.com/ory/client-rust).

So that as a user, I’d just sign up on Weird.one and once signed up I would already have accounts on Revolt-net, Kitsune-net etc.

Related:
- #2 
- https://github.com/commune-os/weird-rs/issues/8
- https://github.com/authifier/authifier/issues/35
- https://github.com/revoltchat/revolt/discussions/103

aumetra:
Closes #123 

This doesn't 100% rely on the OIDC server all the time. Kitsune still creates own OAuth2 tokens but the login is completely handled by the OIDC server.  
If the user does not yet exist on Kitsune, the user is created on-the-fly using the `preferred_username` and `email` fields from the OIDC claim.

The OAuth2 clients for Kitsune never really interact with the OIDC server in any meaningful way.  
The OIDC server is only hit on the `/oauth/authorize` route.


aumetra:
@erlend-sh Do you think this is sufficient for what you thought of? This would only using the OIDC server for login purposes, nothing else (so the token management is completely independent)

erlend-sh:
Sweet! Won’t know for sure until we try, but it sounds like all that’s needed 👌

PR #163, opened 2023-04-06
Initial federation tests for first alpha pre-release

aumetra:


aumetra:
Federation successfully works for `Note` objects against Akkoma. Still investigating why Mastodon won't accept the objects

aumetra:
We had to change how we determine whether to send the HTML page or the ActivityPub JSON representation of the post since Mastodon (for some reason) sends `text/html;q=0.1` in the `Accept` header (which it doesn't for actor requests).  

Since, as far as I know, the HTTP utility libraries don't implement the quality encoding, we simply switched from `.contains(TEXT_HTML.as_ref())` to `!(.contains(APPLICATION_JSON) || .contains(APPLICATION_ACTIVITY_JSON) || .contains(APPLICATION_LD_JSON))`.

Not the prettiest solution, but it works and I really don't want to implement quality encoding right now.

aumetra:
- Likes federate
- Create activities federate
- Media attachments federate

Checking why the delete and undo activities don't work quite yet

PR #165, opened 2023-04-09
ActivityPub type rework

aumetra:
The current layout of the ActivityPub types felt a bit off. This PR changes how they are represented as structs.  
While the new structure does introduce some repeated field names, I think it's still easier to reason about than the previous one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,13 +21,23 @@ We successfully federated with Mastodon on the following functionality:
 - Actors
 - Posts
     - Create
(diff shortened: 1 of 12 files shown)


aumetra:
We still need to add tombstone support because Mastodon sends delete activities in the form of:

```
{
    "type": "Delete",
    "actor": "[whatever]",
    "object": {
        "id": "[whatever]",
        "type": "Tombstone"
    }
}
```

Something in that direction

Issue #203, opened 2023-05-09
Registration form input field's text have wrong colour


perillamint:
## Problem

![image](https://user-images.githubusercontent.com/1622360/237013122-0cad4910-1a81-41cb-ba93-381d2fb424a1.png)

In the current welcome screen, the register form's input field does not have a proper colour and falls back to the default (usually black) colour configuration.

In most cases, this makes it extremely hard for users to interact with the registration form.

## Proposed fix
Add font colour attribute on CSS style to make it use white colour to match other texts in the screen.

aumetra:
Closed with merge of #204

PR #204, opened 2023-05-09
Add white text colour attribute on registration form style

perillamint:
This fixes issue #203 by

* Set up color attribute (white) on the registration forms CSS style.


aumetra:
Thanks!

Issue #221, opened 2023-05-28
Return 401 instead of 404 when attempt to not existing user.


perillamint:
### Pitch

Currently,  Kitsune returns 404 if user tries to sign in not existing user. However, this is not desirable behavior because it emits 404 and this means the destination resource is not available, not authentication failure.

![image](https://github.com/kitsune-soc/kitsune/assets/1622360/d963723d-8b81-4adb-bb46-aebd8534eb4d)

### Proposal

* Properly return 401 by returning `Error::PasswordMismatch` when user does not exist in the database.

PR #222, opened 2023-05-28
Return 401 instead of 404 when user is not found

perillamint:
Fixes issue #221


aumetra (approved):
Good catch! I've been wanting to rework the error handling for a while now, making more specialised errors for each subsection of the code, allowing for easier handling of things like this.

PR #247, opened 2023-06-18
Add link previews

aumetra:
This PR adds link previews where the metadata is fetched via oEmbed.  
It utilises the embed service provided by Lantern chat.

https://github.com/Lantern-chat/embed-service


erlend-sh:
@novacrazy what is the license of embed-service? MIT would be the easiest.

novacrazy:
The `embed-SDK` crate is MIT or Apache 2.0 (it’s specified in Cargo.toml).

The `embed-server` microservice is AGPLv3. Since it’s not something you statically link to then it’s fine to use however.

erlend-sh:
Oh right, individual licenses within the folders. I suggest you add a LICENSE.txt with that info, so that it doesn’t look like the repo s license-less, which makes people think they can’t safely include it in their open source project.

Example: https://github.com/fishfolk/jumpy/blob/main/LICENSE

PR #254, opened 2023-06-30
Improved job scheduler

aumetra:
This PR replaces the current database polling job scheduling approach with something a little more sophisticated.

It delegates the scheduling over to a Redis instance and makes use of Redis Streams and consumer groups to ensure that:

- Jobs are only executed once
- Jobs are re-executed even if the job service crashes

It also supports scheduling jobs to only get run after a specific timestamp. This kind of scheduling is a little inaccurate though, and jobs might be pushed onto the work queue *in a worst case scenario* 5-10 seconds after they are ready.  
This is because we have to poll the scheduled job sorted set and then push the values over onto the stream.  
The operation doing this is transactional though since it is ran via Redis Lua scripting support.

The workers are then just waiting for jobs by blocking on a Redis connection, lowering the strain put on the database.

While Redis is responsible for scheduling jobs, it doesn't contain the job context. The context is stored in the database in the table `job_context`.  
This is mainly because Redis is fully in-memory. Therefore I want to keep the jobs themselves as slim as possible to not fill up memory at some point (important on low-memory machines, which we aim to support as well).

Another improvement this change is supposed to bring is that this should enable to run the job runners as a separate binary. This would enable to possibility to just spin up small VMs that do nothing but run hundreds-of-thousands of jobs in parallel 

## Potential issues

1. We still poll Redis to execute the Lua script every few seconds. Since Redis is single-threaded, no commands can get processed while the script is running
2. Redis needs to be configured by the administrators to provide strong persistence guarantees to proof against crashes, power outages, etc.
3. We introduce yet another service dependency (last release it was PostgreSQL, now it's Redis)


aumetra:
I will probably extract the core job scheduling logic into an own library. Seems like something that might be interesting

PR #268, opened 2023-07-09
Add caddy example file.

SlimeyIceCream:
Added an example file for caddy.

Should be functional.
I'm slightly unsure if the second example I put for using specific certificates is correct because I put it in slightly different than the way I have set it up, but most people won't need to use it anyway.


aumetra:
Thanks!

PR #270, opened 2023-07-11
Add nixos service module to the flake

Nemo157:
I use `deploy-rs` for deploying my machines, this is how I use this module with it, with other ways to configure it's likely to be a little different:

Apply the overlay and module to the system before loading its configuration:
```nix
        kitsune-test = nixpkgs.lib.nixosSystem {
          system = "x86_64-linux";
          modules = [
            ({ ... }: { nixpkgs.overlays = [ kitsune.overlays.default ]; })
            kitsune.nixosModules.default
            kitsune-test/configuration.nix
          ];
        };
```

Configure postgres, kitsune, caddy:

```nix
  services.postgresql = {
    enable = true;
    initialScript = pkgs.writeText "kitsune-init.sql" ''
      CREATE ROLE "kitsune" WITH LOGIN PASSWORD 'kitsune';
      CREATE DATABASE "kitsune" WITH OWNER "kitsune"
        TEMPLATE template0
        LC_COLLATE = "C"
        LC_CTYPE = "C";
    '';
  };

  services.kitsune = {
    enable = true;
    config = {
      instance = {
        name = "Kitsune Test Instance";
        description = "A test of deploying kitsune to nixos";
        registrations-open = false;
      };
      url = {
        scheme = "https";
        domain = "kitsune-test.nemo157.com";
      };
    };
  };

  networking.firewall.allowedTCPPorts = [ 80 443 ];
  services.caddy = {
    enable = true;
    extraConfig = ''
      kitsune-test.example.org {
        reverse_proxy localhost:5000
      }
    '';
  };
```

I decided to hardcode using redis in the module for now, since it needs to do some extra configuration around systemd service dependencies and adding groups to access them.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -119,5 +119,15 @@
           };
         };
       }
-    );
+    ) // {
+      overlays = rec {
+        default = kitsune;
+        kitsune = (import ./overlay.nix self);
+      };
+
+      nixosModules = rec {
+        default = kitsune;
+        kitsune = (import ./module.nix);
+      };
+    };
 }
--- a/flake.nix
+++ b/flake.nix
@@ -23,9 +23,9 @@
           zlib
         ];
         cargoConfig = builtins.fromTOML (builtins.readFile ./.cargo/config.toml);  # TODO: Set the target CPU conditionally
-        cargoToml = builtins.fromTOML (builtins.readFile ./kitsune/Cargo.toml);
+        cargoToml = builtins.fromTOML (builtins.readFile ./Cargo.toml);
         basePackage = {
-          inherit (cargoToml.package) version;
+          inherit (cargoToml.workspace.package) version;
 
           meta = {
             description = "ActivityPub-federated microblogging";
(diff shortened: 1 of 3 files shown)


Nemo157:
It's as done as I can make it with my current understanding of the setup, if you don't notice anything lacking it's good to merge as a base.

The biggest potential issue I see is switching to `x86-64-v3` rather than `native`, that will cause issues for people running CPUs that don't support it. Actually, thinking about it a little more now I just realised this will cause problems for e.g. anyone building on aarch64, I'll remove that commit from this PR and think about another solution for it.

aumetra:
I'd love to merge this right now but I'm planning to look into migrating over to TOML today, so this would just break everything in here.

Is it okay to just keep this open until the TOML migration is done? 

I also offer my help to fix this PR up to work with TOML. I don't just wanna break all your hard work and leave you alone to pick up the pieces.

Nemo157:
Sounds good, I can update this once the TOML changes merge.

aumetra:
I merged the changes to migrate to TOML a few hours ago. I might take a look tomorrow. Pretty exhausted today

Nemo157:
Updated to use toml now, and updated the OP example.

aumetra:
Nice! Thank you so much! I'm gonna go ahead and merge this then

aumetra:
Oh my god. I'm gonna yeet the GitHub actions cache into the sun

PR #271, opened 2023-07-14
Replace Dhall with TOML

### LINKED ISSUE #258: Configuration format/language
Right now we use Dhall which works okay (at least in my opinion, I'd love to get some other opinions on that). It's very Haskell-ish but isn't terribly difficult to get the gist of IMO.

Recently the project we use to get our Dhall implementation in Rust for [changed its status to unmaintained](https://github.com/Nadrieril/dhall-rust).

In theory that isn't too big of a problem, it's not like our builds will break or anything, and the limited scope of Dhall we use for Kitsune will stay available, but there is the question of whether we want to rely on unmaintained crates.

There are multiple options of configuration formats/languages we can consider:

- TOML
- Nickel

(this list gets updated if we get more viable recommendations)

Formats I don't want to use and the reasons:

- Environment variables: No complex expressions (such as structs), so the names will get longer as nesting grows.
    - Sub-point: We will end up with *a lot* of environment variables. Not exactly elegant
- YAML: Advertised as "human-readable" but anything but. An example for YAMLs pitfalls that I like to cite is the "Norway problem"

aumetra:
Dhall has been a great configuration language, unfortunately it has sometimes gotten in the way in the following ways:

- Confusing type errors
- Unfamiliar syntax for people not already a bit used to functional languages, such as Haskell
- Incompatibility with tooling such as Nix

Also `serde-dhall`, which has been pretty much the only way to get Dhall deserialization in Rust, has been deprecated.

---

TOML is a well-known format that is widespread in the Rust ecosystem, with actively developed libraries for deserialization.  
Its format is easy to grasp and it is compatible with Nix. It fulfils pretty much all our requirements.

Dhall configurations should be reasonably easy to convert.

## Changes

- All of the fields and names have been converted to kebab case
- Dhall sum-types have been converted to internally tagged enums
    - Example: `types.Cache.Redis { ... }` -> `[cache] type = "redis" ...`
- Instead of representing absent values with an optional sumtype, they are just omitted
- `request_timeout_sec` was renamed to `request-timeout-secs`
- Number of job workers has been moved to the `job-queue` section and renamed to `num-workers`. It also now enforces to have a value of at least 1 at deserialization 

Closes #258


aumetra:
@Nemo157 does this seem like something that can be worked with or do you have any thoughts on improving this?

https://github.com/kitsune-soc/kitsune/blob/86c7bec924159c5bbf6b8b2a2ab24d4800e4e336/config.example.toml#L1-L42

PR #274, opened 2023-07-18
Do not assume CPU support for aes instructions

Houkime:
Because Intel Core i3 for example does not have them.


aumetra (approved):
Thanks!

PR #278, opened 2023-07-20
WIP: make dev shell self-contained

Houkime:
it now starts nix versions of postgres and redis and all tests pass

https://github.com/kitsune-soc/kitsune/assets/32805122/d4606415-b9f7-47de-a02a-ba0c60d4b7f2

BUT it will conflict with system redis/postgres if they are running. By conflict i mean, it will try to shut them down.
so WIP
And will conflict wwith CI


aumetra:
That's really nice! I also like the custom `PS1`, I never figured out how to make one myself.  
Would you be open to splitting this up into two distinct dev shell definitions? One "full" and one "default", maybe?

The "full" one would pull in the Postgres and Redis dependencies and start them, while the default one doesn't pull these services in.  

That way we could ignore the problem of interfering with system processes (since they would explicitly need to opt into the "full" variant, and we can add a disclaimer to the README contributing section).  
And users who already have a Postgres running in, for example, Docker wouldn't need to boot up yet another instance.

But feel free to add the PS1 to both shells! I will probably end up using the non-full version on my desktop PC and always appreciate a good prompt :D

flake.nix
```diff
@@ -106,11 +106,45 @@
               dhall
```
aumetra, line 106: Also it would be nice if we could just get rid of this in this PR, too. I forgot to remove the Dhall CLI back when I switched configuration formats 

```suggestion

```

aumetra:
What do you think about offloading the work of running the services to `devenv`? https://devenv.sh/

aumetra:
I'm sorry I've been slow recently. I'm gonna merge that one for now.

Thank you so much for your work on it!

PR #280, opened 2023-07-25
Define a fallback value in the build scripts

aumetra:
Nix doesn't copy the `.git` directory into the build environment and therefore the build scripts can't retrieve the Git commit.

For these cases we define a fallback value in the build scripts and set the `VERGEN_GIT_SHA` value to `unknown`


aumetra:
Vergen already defines fallbacks. This was a misinterpretation of build logs on my side

Issue #184, opened 2023-04-16
Follow requests


aumetra:
The basic infrastructure is already there, now we just need to add the Mastodon API endpoint for it (and add the delivery of rejects)

aumetra:
Mastodon API documentation: https://docs.joinmastodon.org/methods/follow_requests/  

Since we already have to implement `Link` headers, we should probably put the encoding logic into a separate function (because theoretically timelines should also have these headers, they just aren't used a lot).

Also don't worry about the equivalent GraphQL function, the whole API is lacking in functionality at the moment since I mainly focussed on the Mastodon API.  
This functionality will just join the rest under a separate task of "getting the GraphQL API up to speed with the Mastodon API"

Zeerooth:
I'll have a shot at this feature, because I see that it's not yet implemented

PR #283, opened 2023-08-08
Implement follow requests

Zeerooth:
Resolves https://github.com/kitsune-soc/kitsune/issues/184

I have also added a "Not Found" page for the frontend router and included public files in the Dockerfiles.

Beware that I have tested it only on a local instances so far, I still need to test the interoperability between instances.


aumetra:
Thank you for giving implementing it a try!

Just a small issue I saw from quickly looking over the code, the Mastodon API docs mention the following about the `max_id` and `since_id`:

> Internal parameter. Use HTTP Link header for pagination.

And I can't see a `Link` header in the handlers. As soon as there is more than one page of requests, the current implementation would probably break with pretty much all clients.

Maybe we could design a nice little struct for `Link` headers that nicely serializes into the correct `HeaderValue` via axum's `IntoResponseParts`?

Zeerooth:
Damn, I wish I could make a pseudo code in rust so easily :D

What you've sent pretty much worked, but I had to change just a few things, like use `String` instead of `&str` for the header key/value (otherwise axum router wouldn't accept it as a return type for the endpoint because of generic types) and use `HeaderValue`

For now I've only added the Link header support for `follow_requests` endpoint, because I'd like to know if it's done okay before I use it anywhere else. From the functional standpoint though it works great and in the same way as the mastodon api (:

aumetra:
It looks generally fine! Just a small nit, I think you can split the types here

```rust
T: IntoIterator<Item = (S, S)>,
S: Display,
```

into two:

```rust
T: IntoIterator<Item = (K, V)>,
K: Display,
V: Display,
```

Which in turn would enable us to do something like this:

```rust
pub type PaginatedJsonResponse<T, const C: usize> = (Option<Link<[(&'static str, String); C]>>, Json<Vec<T>>);
```

That way we wouldn't need to allocate for the names of the keys nor for the container

Zeerooth:
I updated the code and added the Link header to timelines and account statuses

> ```rust
> pub type PaginatedJsonResponse<T, const C: usize> = (Option<Link<[(&'static str, String); C]>>, Json<Vec<T>>);
> ```

Unfortunately this doesn't work with the const parameter, because sometimes we need to only return `Link([next, prev])` and sometimes `Link([prev])`  when we get to the last page of the results (at least that's what mastodon does and I think it's reasonable) so the compiler is always going to complain:
```
expected enum `std::option::Option<pagination::Link<[(&str, std::string::String); 2]>>`
found enum `std::option::Option<pagination::Link<[(&str, std::string::String); 1]>>
```

aumetra:
Ah, yeah. Makes sense. Maybe when I add more optimizations in the future (keyword `smallvec`/`heapless`) that won't be an issue anymore, but as it stands, that's entirely reasonable.

Just one tiny nit before I merge this: Could you split the "accept" and "reject" handlers into their own files, called `accept.rs` and `reject.rs` respectively?  
The rest of the handlers are *somewhat* laid out like they would be with a file-based router, just to make navigating a bit more intuitive.

kitsune/src/http/handler/mastodon/api/v1/timelines/home.rs
```diff
@@ -41,22 +48,57 @@ pub struct GetQuery {
...
+    let mut statuses: Vec<Status> = timeline
         .get_home(get_home)
         .await?
         .and_then(|post| mastodon_mapper.map((&user_data.account, post)))
         .try_collect()
         .await?;
 
-    Ok(Json(statuses))
+    if query.min_id.is_some() {
+        statuses.reverse();
```
aumetra, line 71: Same here

kitsune/src/service/account.rs
```diff
@@ -73,12 +76,40 @@ pub struct GetPosts {
     #[builder(default = API_DEFAULT_LIMIT)]
     limit: usize,
 
-    /// Smallest ID
+    /// Smallest ID, return results starting from this ID
```
aumetra, line 79: Thanks for adding these docs here. I really had a hard time remembering the difference between `min_id` and `since_id` just now.

I really wish the Mastodon docs were better..

aumetra (changes requested):
Just a few comments. It's mixed between regular comments and request changes.

I'm just not good at GitHub's review UI..

kitsune/src/http/handler/mastodon/api/v1/follow_requests.rs
```diff
@@ -0,0 +1,193 @@
...
+    path = "/api/v1/follow_requests/{id}/authorize",
+    security(
+        ("oauth_token" = [])
+    ),
+    responses(
+        (status = 200, description = "Follow request accepted", body = Relationship),
+        (status = 404, description = "No pending follow request from that account ID")
+    ),
+)]
+pub async fn accept(
```
aumetra, line 116: Can we move these into separate modules? To be more in line how the rest of the handlers are laid out?

kitsune/src/http/handler/mastodon/api/v1/follow_requests.rs
```diff
@@ -0,0 +1,193 @@
...
+    path = "/api/v1/follow_requests/{id}/reject",
+    security(
+        ("oauth_token" = [])
+    ),
+    responses(
+        (status = 200, description = "Follow request rejected", body = Relationship),
+        (status = 404, description = "No pending follow request from that account ID")
+    ),
+)]
+pub async fn reject(
```
aumetra, line 158: Same here

aumetra (changes requested):

aumetra:
Honestly, just ignore the `VecDeque` comments. They maybe have an overall lower complexity for inserting but are slower to iterate. I'll probably get rid of them in the other places, too.

aumetra:
If anything is too ambiguous, feel free to ask. I'm also happy to help out and do some of the changes myself

kitsune/src/http/pagination.rs
```diff
@@ -0,0 +1,35 @@
...
+        let value = self
+            .0
+            .into_iter()
+            .map(|(key, value)| format!("<{value}>; rel=\"{key}\""))
+            .collect::<Vec<String>>()
+            .join(", ");
+
+        // as long as we pass valid links this should never panic
+        res.headers_mut()
+            .insert("Link", HeaderValue::from_str(&value).unwrap());
```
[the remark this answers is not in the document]
Zeerooth: I made it so that the header is only appended if it gets parsed correctly
```rust
if let Ok(header_value) = HeaderValue::from_str(&value) {
    res.headers_mut().insert("Link", header_value);
}
```
We should probably add some kind of error log here in the future tho

kitsune/src/service/account.rs
Zeerooth: I agree, I had to dig through github discussions to find that info lmao

kitsune/src/http/handler/mastodon/api/v1/timelines/public.rs
```diff
@@ -68,5 +77,38 @@ pub async fn get(
...
+        let prev = (
+            "prev",
+            format!(
+                "{}/api/v1/timelines/public?limit={}&min_id={}",
+                base_url,
+                query.limit,
+                statuses.first().unwrap().id
+            ),
+        );
+        if statuses.len() >= query.limit && query.limit > 0 {
```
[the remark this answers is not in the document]
Zeerooth: I agree, I moved it into the new `new_link_header()` function

kitsune/src/http/handler/mastodon/api/v1/follow_requests.rs
Zeerooth: Done chief 🫡

kitsune/src/http/handler/mastodon/api/v1/follow_requests.rs
Zeerooth: Done chief 🫡

kitsune/src/http/pagination.rs
```diff
@@ -26,10 +26,46 @@ where
...
+    base_url: &str,
+    get_key: F,
+) -> Option<Link<Vec<(&'static str, String)>>> {
+    if collection.is_empty() {
+        None
+    } else {
+        let next = (
+            "next",
+            format!(
+                "{}/api/v1/timelines/public?limit={}&max_id={}",
```
aumetra, line 50: But this isn't quite right for the `follow_requests` route then, no?

Something that one could do is pass in the URI from the handler.  
So you don't have to type it out manually, you could use something like the `OriginalUri` extractor in the handlers.

kitsune/src/http/pagination.rs
```diff
@@ -0,0 +1,71 @@
...
+        if let Ok(header_value) = HeaderValue::from_str(&value) {
+            res.headers_mut().insert("Link", header_value);
+        }
+
+        Ok(res)
+    }
+}
+
+pub fn new_link_header<I, D: Display, F: Fn(&I) -> D>(
+    collection: &Vec<I>,
```
aumetra, line 39: Using a slice here is more idiomatic 

```suggestion
    collection: &Vec<I>,
    collection: &[I],
```

kitsune/src/http/pagination.rs
```diff
@@ -0,0 +1,71 @@
...
+
+    fn into_response_parts(self, mut res: ResponseParts) -> Result<ResponseParts, Self::Error> {
+        let value = self
+            .0
+            .into_iter()
+            .map(|(key, value)| format!("<{value}>; rel=\"{key}\""))
+            .collect::<Vec<String>>()
+            .join(", ");
+
+        // TODO: log an error when HeaderValue conversion fails
```
aumetra, line 29: Can't we change the associated error type to `InvalidHeaderValue`? Are there trait bounds that aren't satisfied?

kitsune/src/service/account.rs
aumetra: I think the first time I learned the distinction was looking at the docs of Mastodon.py heh

kitsune/src/http/pagination.rs
aumetra: The only concern I have with the approach of passing in the URI is that it's yet *another* parameter and that makes the list of parameters pretty unwieldy, to say the least

kitsune/src/http/pagination.rs
aumetra: Oh, it wants an `IntoResponse` error. I think you can just use our error "super enum" for now. That should implement `From<InvalidHeaderValue>` and `IntoResponse`

Zeerooth: Oh yeah, that works, thanks! I just had to map the error to `http`'s `Error` first

kitsune/src/http/pagination.rs
Zeerooth: You're absolutely right, I fixed that

kitsune/src/http/pagination.rs
Zeerooth: Not changing the route was a big brainfart on my part, sorry, but now I changed it to pass the path as an argument

> The only concern I have with the approach of passing in the URI is that it's yet _another_ parameter and that makes the list of parameters pretty unwieldy, to say the least

That's why I was pretty hesitant at first to make a dedicated function for that, but still, a function with 5 arguments looks better that similar 20 lines of code. 

The only issue now is that by adding `OriginalUri` to the `/api/v1/accounts/statuses` route, there are 8 arguments now which is apparently too much for clippy

aumetra: I just checked the function signature. It looks like we have a duplicated extraction with `auth_data` and `user_data`.  
We should remove one of those anyway. I don't really care particularly about which one we keep, feel free to just choose one.

Zeerooth: Oh right lol, I didn't even notice, but I removed one of these now so it should be ok

kitsune/src/http/pagination.rs
```diff
@@ -0,0 +1,76 @@
...
+        res.headers_mut().insert(
+            "Link",
+            HeaderValue::from_str(&value).map_err(HttpError::from)?,
+        );
+
+        Ok(res)
+    }
+}
+
+pub fn new_link_header<I, D: Display, F: Fn(&I) -> D>(
```
aumetra, line 40: A silly idea I had.. what about making this an associated function of `Link` with the name `from_collection`.

And the parameter it takes is a struct annotated with the `TypedBuilder` macro to have the compile-time enforced builder pattern there.

Would probably provide a more readable code with less parameters

aumetra: Something like 

```rust
#[derive(TypedBuilder)]
struct CollectionParams {
    [fields here]
}
```

And the usage would be

```rust
let params = CollectionParams::builder()
    .base_uri(..)
    .build();

Link::from_collection(params)
```

Zeerooth:
I can't get this to work
```rust
pub type PaginatedJsonResponse<T> = (
    Option<Link<Either<[(&'static str, String); 1], [(&'static str, String); 2]>>>,
    Json<Vec<T>>,
);

pub struct Link<T>(pub T);

impl<T, K, V> IntoResponseParts for Link<T>
where
    T: IntoIterator<Item = (K, V)>,
    K: Display,
    V: Display,
{
    type Error = Error;

    fn into_response_parts(self, mut res: ResponseParts) -> Result<ResponseParts, Self::Error> {
        let value = self
            .0
            .into_iter()
            .map(|(key, value)| format!("<{value}>; rel=\"{key}\""))
            .collect::<Vec<String>>()
            .join(", ");

        res.headers_mut().insert(
            "Link",
            HeaderValue::from_str(&value).map_err(HttpError::from)?,
        );

        Ok(res)
    }
}
```

```rust
error[E0277]: `[(&str, std::string::String); 1]` is not an iterator
  --> kitsune/src/http/handler/mastodon/api/v1/follow_requests/mod.rs:61:6
   |
61 | ) -> Result<PaginatedJsonResponse<Account>> {
   |      ^^^^^^ `[(&str, std::string::String); 1]` is not an iterator; try calling `.into_iter()` or `.iter()`
   |
   = help: the trait `std::iter::Iterator` is not implemented for `[(&str, std::string::String); 1]`
   = help: the following other types implement trait `askama_axum::IntoResponse`:
             std::result::Result<T, E>
             std::result::Result<T, axum::response::ErrorResponse>
   = note: required for `either::Either<[(&str, std::string::String); 1], [(&str, std::string::String); 2]>` to implement `std::iter::Iterator`
note: required for `Link<Either<[(&str, String); 1], [(&str, String); 2]>>` to implement `axum::response::IntoResponseParts`
  --> kitsune/src/http/pagination.rs:19:15
   |
19 | impl<T, K, V> IntoResponseParts for Link<T>
   |               ^^^^^^^^^^^^^^^^^     ^^^^^^^
20 | where
21 |     T: IntoIterator<Item = (K, V)>,
   |                     ------------- unsatisfied trait bound introduced here
```

yet `[(&str, std::string::String); 1]` implements iterator and so should either, because it states in the docs that `Either<L, R> is an iterator if both L and R are iterators.` so I have no idea why the trait bound doesn't work.
Either way, I feel like using Either in most cases, as well as here is a weird hack so I don't know if it's worth the effort.

kitsune/src/http/pagination.rs
Zeerooth: Builder patterns are nice, but I feel that here it's redundant. Maybe if we had some default arguments to explicitly provide or like 8 args making clippy scream at our code then yes, but not in this case.

aumetra: Yeah, fair. I still feel like making it an associated function of "Link" would be nice here though

aumetra:
Okay, yeah. That's weird (especially considering the blanket impl they have for `IntoIterator`)

But yeah, feel free to just not do it. It's really not worth the effort

kitsune/src/http/pagination.rs
Zeerooth: I agree 100%, sorry for not doing it before, but I made it into an associated fn now :)

aumetra:
That looks really good now. Thanks for adding this!

aumetra (approved):

PR #293, opened 2023-08-13
Language detection for posts

### LINKED ISSUE #252: Language detection of posts
For better full-text search and eventual support for translators.

We can either store the language the user submits to us via the endpoint, but there we have the possibility of someone setting the language wrong by accident.

We could probably in the beginning base our detection on [whichlang](https://github.com/quickwit-oss/whichlang), which has a pretty high accuracy

aumetra:
Auto-detects the language of posts via `whatlang` and automatically tokenizes the posts with the language-specific tokenizer inside Postgres (if installed).

To ensure that we only use tokenizers the user has actually installed, we use an ugly `pg_catalog` hack at start-up.  
We generate a mapping function from ISO codes to the english language name (tokenizers are pretty much always named after their english names) for all installed tokenizers.

Since the function is marked as immutable, the PostgreSQL optimizer should do its damnedest to optimize the function into just a table lookup.

(thanks `pg_catalog.pg_ts_config` lol)

Closes #252


aumetra:
Making this into an PostgreSQL enum will be difficult. We have the problem that Diesel wants to generate the enum definition from the schema, but the enum definition is dynamically generated at start-up.

I feel like the easiest solution for now is to just store the ISO code as a `TEXT` or `VARCHAR(3)` (shouldn't actually make a difference. IIRC Postgres uses the same internal representation for both types).

aumetra:
The MVP here is finished. The full text search works and is splitting words based on their content language and is automatically detecting newly installed dictionaries (I tested it with a Korean tokenizer).

Needs some clean-up and some ideas about how to store the ISO code properly, though.

---

Another To-Do: Send out the language information via the API.

aumetra:
Just tested it against Russian in Cyrillic script and it worked well. Tokenized the posts correctly and the search worked, too.

aumetra:
I think I found a way to make this work with an enum, meaning more efficient representation in-database (one integer instead of two integers + whatever size the string is)

Basically, we create an empty enum inside the migrations and on start-up we run the following for each supported ISO language:

```sql
ALTER TYPE iso_language_code ADD VALUE IF NOT EXISTS '{iso code}';
```

And for deserialization, we just define a small transparent wrapper type around `whatlang::Lang` and use its `Lang::from_code` to deserialize it inside the `FromSql` handler (since in Postgres, you get the defined string value of the enum sent over the wire)

```rust
Self(whatlang::Lang::from_code(unsafe { str::from_utf8_unchecked(bytes.as_bytes()) })?)
```

aumetra:
And just like that we use PostgreSQL enums. Nice.  
This should have a more efficient on-disk representation of four bytes.

Now I'm thinking about maybe using a crate like `isolang` to stay in line with the actual ISO standard and not with languages that `whatlang` happens to support

aumetra:
Also, something I wanted to do for a long time is just squash down the two rows I have per database entry for the full-text search (`content_ts` and `subject_ts`, etc.) into a single field, via ts vector concatenation with `setweight`

aumetra:
So, now that that's cleaned up, time to think about the ISO code situation

aumetra:
Well.. now we have a complete copy of all ISO codes. It's around 8000 codes. Seems a little.. excessive.

aumetra:
We are now only creating an enum of the most commonly used languages, as defined by ISO-639-1.  
Those are ~200 languages which is, in my uneducated opinion, exhaustive enough.

aumetra:
Now that we are using `isolang` anyway I'm torn on whether we should actually use `whatlang` or `whichlang`.

While `whatlang` has support for more languages, `whichlang` is faster and slightly more accurate.  
Maybe we can make this a compile-time feature flag?

aumetra:
`lingua` adds a heavy cost in both binary size and memory usage, so I'm gonna keep it disabled by default.  
But people who want to opt into that additional cost can do so

aumetra:
Okay, but do we even need `lingua` support? That seems like something not a lot of people would even use. And with the backend toggle we have right now, it would be trivial to add more backends.

I think I'll remove support for it from this MVP until someone asks for it

PR #296, opened 2023-08-15
test: Use mocked HTTP clients

### LINKED ISSUE #294: Mock the HTTP client in tests
Some of current test code makes HTTP requests to remote servers, for example:

https://github.com/kitsune-soc/kitsune/blob/a098f0a1647cff01d2b8b110a2b942ed73bcddea/kitsune/src/activitypub/fetcher.rs#L300-L303

But this kind of test is not quite reproducible since it would break if the server went down or you removed the account. Also, it might be slow and possibly overload the server, and you'll need to find another server in the wild if you are going to test against a feature the current target server doesn't support (especially, #285 is in my mind).

Fortunately, Kitsune is using Tower's middleware abstraction for its HTTP client, which is easy to replace with a mock client. That can be achieved by adding a new finalizer method to `kitsune_http_client::ClientBuilder` that takes a mock HTTP client to wrap. With the new method (let's call it `build_with_client` for now) the test code will look like the following pseudo code:

```rust
let mock_client = tower::service_fn(|req| async move {
    let res = create_mock_response();
    Ok(res)
});
let client = Client::builder()/* ... */.build_with_client(mock_client);
let fetcher = Fetcher::builder().client(client)/* ... */.build();

let user = fetcher
    .fetch_actor("https://corteximplant.com/users/0x0".into())
    .await
    .expect("Fetch actor");
```

If this idea looks good to you, I'd be happy to work on a pull request!

tesaguri:
Closes #294.

This adds a new `kitsune_http_client::ClientBuilder::service` finalizer method and uses it to replace the underlying HTTP clients with mocked ones in tests (this also adds `kitsune::webfinger::Webfinger::with_client` constructor to enable the mocking).

I named the `service` method after [`tower::ServiceBuilder::service`] method, which is used in the same way as the new method. But there are alternative candidates like `client` (since it mainly takes a client service) and `build_with_service` (which emphasizes that it's a finalizer method akin to `build`).

The method turned out to be exclusively used in combination with `tower::service_fn` in the tests. I tried to keep the API addition as minimal as possible for starters, but If deemed worth, we could add a shorthand method such as `ClientBuilder::service_fn` (similar to [`tower::ServiceBuilder::service_fn`]) and `Client::service_fn`.

[`tower::ServiceBuilder::service`]: <https://docs.rs/tower/0.4.13/tower/struct.ServiceBuilder.html#method.service>
[`tower::ServiceBuilder::service_fn`]: <https://docs.rs/tower/0.4.13/tower/struct.ServiceBuilder.html#method.service_fn>


aumetra (approved):
This looks really good! Thanks for the PR!

About the potential future API additions, I would merge this as an MVP where we can do subsequent changes, such as adding `Client::service_fn` or `Client::default_with_service`, in the future

Issue #285, opened 2023-08-11
Custom WebFinger domain (à la Mastodon's `LOCAL_DOMAIN`)


tesaguri:
Mastodon uses (misleadingly named IMO) [`LOCAL_DOMAIN`] environment variable value to set the domain part of canonical WebFinger `subject`s of the server. It would be great if Kitsune supported a similar configuration.

[`LOCAL_DOMAIN`]: <https://docs.joinmastodon.org/admin/config/#local_domain>

Suppose Alice is running a  Mastodon server at `social.alice.example` domain. Naively, Alice's handle name would be `@alice@social.alice.example`, but why bother with the `social.` subdomain? In the world of e-mail, Alice can run an SMTP server at `smtp.alice.example` and receive mails as `alice@alice.example` by setting appropriate DNS records. `LOCAL_DOMAIN` works just like that: Alice's account will be recognized as `@alice@alice.example` by setting `LOCAL_DOMAIN` to `alice.example` (and setting up appropriate HTTP redirections in `alice.example` domain), while the actual Mastodon server is running at `social.alice.example`.

[`social.treehouse.systems`](<https://social.treehouse.systems>) is an example of such a Mastodon server. When you look up <https://social.treehouse.systems/@ariadne> in another Mastodon server, the account will be shown as `@ariadne@treehouse.systems` instead of `@ariadne@social.treehouse.systems`. But how does the other Mastodon server know that `@ariadne@treehouse.systems` is the canonical handle?

Let's follow the discovery process by hand. First, fetch the Activity Streams object at <https://social.treehouse.systems/@ariadne>, which tells us that `ariadne` is the actor's `preferredUsername`. Then, perform a WebFinger query <https://social.treehouse.systems/.well-known/webfinger?resource=acct:ariadne@social.treehouse.systems>, and we get a JRD document with its `subject` field being `acct:ariadne@treehouse.systems`. But since the domain part of the `subject` is different from the ActivityPub actor's domain, we need to verify with `treehouse.systems` by performing another WebFinger query <https://treehouse.systems/.well-known/webfinger?resource=acct:ariadne@treehouse.systems>, just in case `social.treehouse.systems` is making a bogus claim to the authority over the other domain. If the second WebFinger query result agrees with the first one (or simply redirects back to <https://social.treehouse.systems/.well-known/webfinger?resource=acct:ariadne@treehouse.systems>), then we can finally confirm that `acct:ariadne@treehouse.systems` is the canonical subject!

[...]

PR #297, opened 2023-08-17
Add `webfinger-domain` configuration

tesaguri:
This adds ~~`instance.webfinger_domain`~~ `instance.webfinger-domain` configuration option to set a custom domain name for the `subject` of local WebFinger resources. See #285 for the background.

The configuration will look like the following:

```toml
[instance]
# ...
webfinger-domain = "example.com"

[url]
scheme = "https"
domain = "social.example.com"
```

This example will run the server at `social.example.com` and make the WebFinger handler resolve local actors as `acct:{username}@example.com` instead of `acct:{username}@social.example.com`.

`GET /api/v1/instance` of Mastodon API and its counterpart in GraphQL API is updated to use the new configuration value to make it align with Mastodon's implementation (though Mastodon doesn't provide a GraphQL API. But since the old value can trivially be known from the endpoint URL, I think the change makes sense for the GraphQL API as well).

## The name?

The configuration name is up for bikeshedding. There are few alternatives in my mind:

- `local-domain` (consistent with Mastodon, arguably the most popular implementaion. Might be confusing though)
- `webfinger-host`(might be more accurate since it may contain a port in addition to the domain, but is inconsistent with the existing `url.domain` config)
- `acct-domain` or `acct-host` (in reference to the `acct` field of Mastodon API)
- `display-domain` (adopted from [Takahē][takahe-doc-domains]. Personally I like it too because of consistency with the term "display name" for `@user` handle name)
- `account-domain` (adopted from [GoToSocial][gts-doc-split-domain])

[takahe-doc-domains]: <https://docs.jointakahe.org/en/stable/domains/>
[gts-doc-split-domain]: <https://docs.gotosocial.org/en/latest/advanced/host-account-domain/#configuration>

I chose the name `webfinger-domain` because it is consitent with `url.domain` and also helpfully reminds the administrator that it's about WebFinger, for which they need to set up an HTTP redirection at the pointed domain.

### DIFF
--- a/crates/kitsune-type/src/webfinger.rs
+++ b/crates/kitsune-type/src/webfinger.rs
@@ -1,7 +1,7 @@
 use serde::{Deserialize, Serialize};
 use utoipa::ToSchema;
 
-#[derive(Deserialize, Serialize, ToSchema)]
+#[derive(Debug, Deserialize, Serialize, ToSchema)]
 pub struct Link {
     pub rel: String,
     pub r#type: Option<String>,
--- a/kitsune/src/config.rs
+++ b/kitsune/src/config.rs
@@ -52,6 +52,7 @@ pub enum FederationFilterConfiguration {
 pub struct InstanceConfiguration {
     pub name: SmolStr,
     pub description: SmolStr,
+    pub webfinger_domain: Option<SmolStr>,
     pub character_limit: usize,
     pub federation_filter: FederationFilterConfiguration,
     pub registrations_open: bool,
--- a/kitsune/src/http/graphql/query/instance.rs
+++ b/kitsune/src/http/graphql/query/instance.rs
@@ -16,7 +16,7 @@ impl InstanceQuery {
         let url_service = &state.service.url;
 
         let description = instance_service.description().into();
-        let domain = url_service.domain().into();
+        let domain = url_service.webfinger_domain().into();
         let local_post_count = instance_service.local_post_count().await?;
         let name = instance_service.name().into();
         let registrations_open = instance_service.registrations_open();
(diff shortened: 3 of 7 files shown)


aumetra:
Sorry about the merge that messed things up. You should probably be able to pretty easily rebase though.  
It was just me moving around directories.

I'm gonna have a look at it after work.

aumetra:
Also just as a small pre-nit, can you rename the configuration option to be written in kebab case to be in line with the other options?

`webfinger_domain` -> `webfinger-domain`

tesaguri:
Rebased onto master.

> Edit: Never mind, should be handled already by the `serde` annotation

Yeah, I've also forgotten about `#[serde(rename_all = "kebab-case")]` and wrote a wrong description. Updated it as well.

aumetra:
How does our Webfinger resolver interact with these display domains? It doesn't respect the display domain IIRC, so that should probably be patched, too.  
The `domain` field in our database, that we use to distinguish servers, is filled by parsing the actor URL (which will pretty much always be to the canonical domain; i.e. `social.example.org` and not `example.org`).

In general we will have a harder time of supporting this completely since Mastodon *really* likes the Webfinger protocol and pretty much inseparably joins ActivityPub and Webfinger.  
It will fetch a user by its URL, use the username from the actor to send a Webfinger request, to obtain the display domain.

Kitsune doesn't do this. We see Webfinger as a QoL improvement that is stitched onto ActivityPub. Kitsune works fine without Webfinger.

I'm not entirely sure how it will interact with all other parts if we just patch an option into `FetchOpts` for the `Fetcher` structure.

tesaguri:
> so that should probably be patched, too.

Do you mean that should be done in this PR? I thought that that can be done separately in a follow-up PR (which I meant to track as Todo 2 of #285).

aumetra:
Oh, sorry, yeah makes sense! I'll remove the issue association with this PR then.

Yeah, let's keep these two things in separate PRs

aumetra:
Okay, so, I fucked up a PC upgrade and ordered wrong components. Now I don't have a computer right now.

Gonna review it when I manage to get it running again

aumetra (approved):
Thanks for the PR! This looks really good!

Issue #244, opened 2023-06-15
Captcha support


aumetra:
To prevent problems like Mastodon has with a bunch of bots signing up, it would be useful to support some kind of Captcha (or proof-of-work; yes, technically not a Captcha) system.

The best option at the moment seems to be [mCaptcha](https://mcaptcha.org/). It's open-source, self-hostable, and since it's PoW-based it shouldn't have issues with accessibility (but we should investigate this before settling on this one).  

Other options include:

- [hCaptcha](https://www.hcaptcha.com/): Not sure about this one. They claim to be privacy-friendly but I wouldn't put my money on that claim tbh
- reCaptcha: No, never. I don't think I need to elaborate on this one. This is at best a desperate last resort of all other services in the world would stop existing

Zeerooth:
Another viable captchas to consider are:
- [captcha crate](https://docs.rs/captcha/latest/captcha/) - This is the one that lemmy uses, but it's quite easy to break and annoying to use as it requires solving a challenge every time as opposed to other proof of work or DDoS detection captchas. Also no API, just pure image generation with the 
- [Cloudflare Turnstile](https://www.cloudflare.com/products/turnstile/) - Free, closed source, so not really to be trusted, but they claim that they don't harvest any data for ad targeting and the like. The question is if these statements can be trusted?

Anyhow, for me it seems that mCaptcha is the best choice for now, so it'd be the best to implement it first. However, I also think that it'd be useful to have an interface and config variables that allow server admins to choose their preferred captcha provider. What do you think?

aumetra:
Yeah, I think we should provide an interface similar to what we have for the storage providers.  
Abstraction crate with the name `kitsune-captcha` and then an internally tagged union via a `type` field inside the configuration.

I think a good line-up for an MVP here would be support for `mCaptcha` and `hCaptcha`.

mCaptcha for more privacy conscious folks and hCaptcha for most others.

---

About the `captcha` crate: I have accessibility concerns. Vision impaired folks might not be able to get through the registration process with that crate and there is no alternative, such as an audio challenge.

hCaptcha offers "accessibility cookies" which is, presumably, a manual application process, but they at least have a solution.  
mCaptcha is fully based on proof-of-work, just requiring computation resources, so no accessibility concerns there.

Zeerooth:
> I'm gonna check when I get the time, but I'm assuming they either have an audio challenge or an alternative bypass method.

[...]

PR #299, opened 2023-08-17
Add captcha

Zeerooth:
Resolves https://github.com/kitsune-soc/kitsune/issues/244

hCaptcha and mCaptcha should work for registration now with this PR, but I still want to iron out the code a bit and maybe add some documentation.

mCaptcha doesn't have a vue library available, so I had to use the native version and plug in an event listener for the messages it emits. Not ideal, but I don't see a better way to do it.


aumetra:
Can you maybe run `yarn format` over the frontend code? I think there's some unformatted things in there

Zeerooth:
> Can you maybe run `yarn format` over the frontend code? I think there's some unformatted things in there

Sure, I didn't even know such a yarn command existed :D

aumetra:
Yeah, we aliased `prettier` to it. Makes the code pretty :>

I'll have a look at the PR as soon as my PC is back in action

crates/kitsune-captcha/src/mcaptcha.rs
```diff
@@ -0,0 +1,56 @@
...
+            .token(token.to_string())
+            .key(self.site_key.to_string())
+            .build();
+        let body = simd_json::to_string(&body)?;
+        let request = Request::post(self.verify_url.clone())
+            .header("Content-Type", "application/json")
+            .header("Accept", "application/json")
+            .body(body.into())?;
+        let response = self.client.execute(request).await?;
+        let mut response_bytes = response.text().await?.into_bytes();
```
aumetra, line 48: We do have a `.json()` method. The API is pretty close to `reqwest`'s

https://github.com/kitsune-soc/kitsune/blob/870f9688b7a297a85988141d79a06bdb322d3656/crates/kitsune-http-client/src/lib.rs#L369-L375

kitsune-fe/src/components/Captcha.vue
```diff
@@ -0,0 +1,84 @@
...
+    <component is="script" type="text/javascript" charset="utf-8">
+      function mcaptcha__load() { let config = { widgetLink: new URL("{{
+        sitekey
+      }}"), }; new mcaptchaGlue.default(config); }
+    </component>
+  </div>
+</template>
+
+<script setup lang="ts">
+  import VueHcaptcha from '@hcaptcha/vue3-hcaptcha';
```
aumetra, line 34: I'm pretty sure this isn't lazy-loading the component. This means the Captcha componment is always loading the hCaptcha code, regardless of whether it is actually used.

https://vuejs.org/guide/components/async.html#basic-usage

kitsune-fe/src/components/Captcha.vue
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  function onError(err: string) {
+    captchaState.token = '';
+    captchaState.error = err;
+  }
+
+  // this is the only to capture mCaptcha token
+  onMounted(() => {
+    window.addEventListener('message', (e) => {
```
aumetra, line 70: I'm pretty sure our linter configuration will complain about this.

Check the output of `yarn lint`

public/mcaptcha-glue.js
```diff
@@ -0,0 +1 @@
...",t.ID="mcaptcha__widget-container";t.default=function(e){var n=this;this.setToken=function(e){return n.inputElement.value=e},this.receiver=new r.default(e,this.setToken),this.receiver.listen();var i=document.getElementById(t.ID);if(null==i)throw new Error("Element ".concat(t.ID,"'s parent element is undefined"));this.inputElement=document.createElement("input"),this.inputElement.id=t.INPUT_NAME,this.inputElement.name=t.INPUT_NAME,this.inputElement.hidden=!0,this.inputElement.required=!0,i.appendChild(this.inputElement);var o="mcaptcha-widget__iframe",a=document.createElement("iframe");a.title="mCaptcha",a.src=this.receiver.widgetLink.toString(),a.ariaRoleDescription="presentation",a.name=o,a.id=o,a.scrolling="no";try{a.sandbox="allow-same-origin allow-scripts allow-popups"}catch(e){try{a.sandbox.add("allow-same-origin"),a.sandbox.add("allow-scripts"),a.sandbox.add("allow-popups")}catch(e){a.setAttribute("sandbox","allow-same-origin allow-scripts allow-popups")}}a.width="100%",a.height="100%",a.frameBorder="0",i.appendChild(a)}}},t={},function n(i){var o=t[i];if(void 0!==o)return o.exports;var r=t[i]={exports:{}};return e[i].call(r.exports,r,r.exports,n),r.exports}(607);var e,t}));
```
aumetra, line 1: Do we have to ship this blob ourselves or is there an NPM package we could use?

crates/kitsune-captcha/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+name = "kitsune-captcha"
+version.workspace = true
+edition = "2021"
+
+[dependencies]
+async-trait = "0.1.73"
+enum_dispatch = "0.3.12"
+http = "0.2.9"
+http-body = "0.4.5"
+hyper = { version = "0.14.27", features = [
```
aumetra, line 11: I think there are a few dependencies you don't actually use declared here

crates/kitsune-captcha/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+    "client",
+    "deprecated",
+    "http1",
+    "http2",
+    "runtime",
+] }
+hyper-rustls = { version = "0.24.1", features = ["http2"] }
+kitsune-http-client = { path = "../kitsune-http-client" }
+serde = { version = "1.0.183", features = ["derive"] }
+serde_qs = "0.12"
```
aumetra, line 21: Do we need support for nested URL parameters? If not, please replace this with `serde_urlencoded`

aumetra (changes requested):

public/mcaptcha-glue.js
aumetra: Since we aren't using SSR, I'm pretty sure you could run the code from the `@mcaptcha/vanilla-glue` NPM package inside the setup tag, using the `onMounted()` lifecycle hook

kitsune-fe/src/components/Captcha.vue
Zeerooth: It did not, actually. I fixed the other linter complaints though, except for one.
If you apply this suggestion:
```
/src/components/AuthForms.vue
  99:22  warning  Replace `username:·$username,·email:·$email,·password:·$password,·captchaToken:·$captchaToken` with `⏎··········username:·$username⏎··········email:·$email⏎··········password:·$password⏎··········captchaToken:·$captchaToken⏎········`  prettier/prettier
```
You get an error:
```
src/components/AuthForms.vue:92:5 - error TS2345: Argument of type 'unknown' is not assignable to parameter of type 'string | DocumentNode | TypedDocumentNode<any, AnyVariables>'.

 92     graphql(`
        ~~~~~~~~~
 93       mutation registerUser(
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
...
107       }
    ~~~~~~~
108     `),
    ~~~~~~
```

aumetra: It uses an interesting strategy of selecting generated TypeScript types by indexing a map for the correct type, where the GraphQL query is the key.  
That's why you can't change the query without regenerating the code.

crates/kitsune-captcha/src/hcaptcha.rs
```diff
@@ -0,0 +1,56 @@
...
+    pub secret_key: String,
+    #[builder(default = Client::builder().build())]
+    client: Client,
+}
+
+#[derive(Serialize, Deserialize, TypedBuilder)]
+struct Body {
+    secret: String,
+    response: String,
+    #[builder(default = None)]
```
aumetra, line 22: You don't have to explicitly declare a value here. When omitted, it will just call the `Default` trait impl

Same with the `client` field of the `Captcha` struct.

Zeerooth: Ah, that's good to know :D

aumetra (approved):
Looks really good now! Thanks!

aumetra:
Also about the missing Vue library, I'm thinking about writing one and just contributing it to them.  
They already have React and Svelte libraries, so it would fit in well, I think.

Zeerooth:
> Also about the missing Vue library, I'm thinking about writing one and just contributing it to them. They already have React and Svelte libraries, so it would fit in well, I think.

That would really be awesome! I was also considering this but I lack the expertise and nerves to deal with vue for now lol

PR #303, opened 2023-08-23
Wrapper around the Postgres connection pool

aumetra:
Right now we use the `deadpool` connection pool directly. This works fine but has some downsides.

One of them is that, since it is guard-based, it's harder to keep track where a connection is going to be returned to the pool.

This PR adds a wrapper with the two functions `with_connection` and `with_transaction`. These two functions accept a closure that takes a connection as a parameter that returns a future.

That way we can easier track scopes in which database connections are borrowed from the pool, and lets us define tighter scopes easier (and implicitly discourages long-term ownage of connections by forcing another level of indentation).

### DIFF
--- a/crates/kitsune-db/src/pool.rs
+++ b/crates/kitsune-db/src/pool.rs
@@ -1,9 +1,19 @@
 use diesel_async::{
-    pooled_connection::deadpool::{Object, Pool, PoolError},
+    pooled_connection::deadpool::{Object, Pool, PoolError as DeadpoolError},
     scoped_futures::ScopedBoxFuture,
     AsyncConnection, AsyncPgConnection,
 };
 use std::future::Future;
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+pub enum PoolError<E> {
+    #[error(transparent)]
+    Pool(#[from] DeadpoolError),
+
+    #[error("{0}")]
+    User(E),
+}
 
 #[derive(Clone)]
 pub struct PgPool {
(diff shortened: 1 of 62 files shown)


crates/kitsune-db/src/pool.rs
```diff
@@ -0,0 +1,47 @@
...
+use std::future::Future;
+
+#[derive(Clone)]
+pub struct PgPool {
+    inner: Pool<AsyncPgConnection>,
+}
+
+impl PgPool {
+    /// Run the code inside a context with a database connection
+    pub async fn with_connection<F, Fut, T, E>(&self, func: F) -> Result<T, E>
```
aumetra, line 15: I'm not happy with this API. It always requires a compatible error type.  
That just sucks because every invokation has to get its error mapped into a compatible type, adding a bunch of `.map_err(Error::from)` repetiton

aumetra: Found a nice solution for this. Created a new error enum that wraps the pool error and a user supplied error, we then just need a single `From` impl on the error type

crates/kitsune-db/src/pool.rs
```diff
@@ -0,0 +1,56 @@
...
+    User(E),
+}
+
+#[derive(Clone)]
+pub struct PgPool {
+    inner: Pool<AsyncPgConnection>,
+}
+
+impl PgPool {
+    /// Run the code inside a context with a database connection
```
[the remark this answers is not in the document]
aumetra: Added some notes about the intent as a struct-level documentation

PR #306, opened 2023-08-24
Update, boost and unboost a status using mastodon api

Zeerooth:


aumetra:
About the index update on the search service we have in-tree, we use tantivy so our indices are immutable.  

The implementation would just delete and add the document.

aumetra:
The rebase of this whole thing shouldn't be *too* bad. Just the way the database access works changed on main

Zeerooth:
> About the index update on the search service we have in-tree, we use tantivy so our indices are immutable.
> 
> The implementation would just delete and add the document.

Is it a bad thing? Would you recommend doing something else instead? I guess I could just remove and then add the document manually without having a dedicated method then

aumetra:
> Is it a bad thing? Would you recommend doing something else instead? I guess I could just remove and then add the document manually without having a dedicated method then

No no, I wouldn't say you should do a different thing. Different search engines might have different optimizations. Maybe a search engine we would consider in the future has some sort of optimization for replacing things.

And even if they don't, a dedicated update method conveys intent better (`update_in_index` makies the intent immediately clear while successive `remove` and `add` calls might need comments to explain). 

The single additional allocation for boxing the future is worth the clearer intent.
(I mean, it would really be zero cost as soon as async traits hit stable)

Zeerooth:
> > Is it a bad thing? Would you recommend doing something else instead? I guess I could just remove and then add the document manually without having a dedicated method then
> 
> No no, I wouldn't say you should do a different thing. Different search engines might have different optimizations. Maybe a search engine we would consider in the future has some sort of optimization for replacing things.
> 
> And even if they don't, a dedicated update method conveys intent better (`update_in_index` makies the intent immediately clear while successive `remove` and `add` calls might need comments to explain).
> 
> The single additional allocation for boxing the future is worth the clearer intent. (I mean, it would really be zero cost as soon as async traits hit stable)

True, meilisearch for example can make use of add_or_update which should be faster than removing and adding.

I've added a simple implementation for kitsune seach now (unfortunately with one clone)

kitsune/Cargo.toml
```diff
@@ -51,6 +51,11 @@ http = "0.2.9"
 human-panic = "1.1.5"
 hyper = { version = "0.14.27", features = ["deprecated"] }
 iso8601-timestamp = "0.2.11"
+isolang = { version = "2.3.0", features = [
```
aumetra, line 54: Is this dependency used somewhere?

kitsune/src/service/post.rs
```diff
@@ -357,11 +457,261 @@ impl PostService {
...
+
+        Ok(post)
+    }
+
+    /// Boost a post
+    ///
+    /// # Panics
+    ///
+    /// This should never panic. If it does, create a bug report.
+    pub async fn boost(&self, boost_post: BoostPost) -> Result<Post> {
```
aumetra, line 580: We usually use the term "repost" throughout the codebase. "Boost" is a bit Mastodon-specific.

Even Mastodon calls it just "repost" in its data schema IIRC

kitsune/src/service/post.rs
```diff
@@ -207,16 +303,20 @@ impl PostService {
...
         } else {
             create_post.content
         };
         content.clean_html();
-        let content_lang = kitsune_language::detect_language(DetectionBackend::default(), &content);
+
+        let detect_language =
+            |s: &str| kitsune_language::detect_language(DetectionBackend::default(), s);
+        let content_lang = create_post.language.map_or_else(
+            || detect_language(&content),
```
aumetra, line 317: I think you can just inline the closure here. The path `kitsune_language::detect_language` makes it pretty clear what it does here

kitsune/src/service/post.rs
```diff
@@ -357,11 +457,261 @@ impl PostService {
...
     }
 
+    /// Update a post and deliver the update
+    ///
+    /// # Panics
+    ///
+    /// This should never ever panic. If it does, create a bug report.
+    pub async fn update(&self, update_post: UpdatePost) -> Result<Post> {
+        if let Some(content) = update_post.content.as_ref() {
+            if content.chars().count() > self.instance_service.character_limit() {
```
aumetra, line 472: Can you just add a `// TODO(aumetra)` here mentioning a migration to `garde` for character limit enforcing?

I wanted to do it for a while now but always forgot

aumetra (changes requested):

kitsune/src/service/post.rs
aumetra: You can mention #307 here

kitsune/src/service/post.rs
aumetra: Just another small ping because of this.  
Would you prefer to keep it inside a separate closure? Because I personally don't see an advantage

Zeerooth: The thing is that here it's used twice, so I kept it in. I only removed it from `update` because there it was only called once so there wasn't a point, but I think here it makes things a bit more concise.

aumetra: Oh, I completely overlooked the second usage. You're right.

(been a while since I was last on my laptop, damn stuff is so much smaller)

aumetra (approved):

aumetra:
Thanks for the additions!

PR #328, opened 2023-08-30
Add notifications

Zeerooth:
This is not yet a final version, I still need to test most of the notification types, whether they respect the preferences and so on, pretty up the queries, but most of the work is done and mastodon endpoint should work.

Some information about the choices I made during the implementation:
- Yes, I know that the `NewNotificationBuilder` impl of custom build method is ugly as hell, but first of all, [it's the recommended way to do it](https://github.com/idanarye/rust-typed-builder/blob/master/examples/complicate_build.rs) by the `typed-builder` crate and second of all, it allows us to ensure that we always provide all the required fields for the notifications table for the given notification type.
- Speaking of which, in the end I decided to make separate columns for foreign keys to account, post tables. It allows us to keep REFERENCES and make unique constraints, which simply makes it superior to having a single, loose activity_id column.  It also allows for greater flexibility in the future if we choose to reference multiple activities.
- I threw around the idea of having a worker that cleans up old notifications, I still think it's necessary, but I'll make an issue for it after we merge this because this PR is large enough as it is, I don't want to bloat it even more.

That should be it, I'm mostly proud of this PR :)


tesaguri (commented):
A big step forward for feature-completeness! I'm excited!

Here's some quick suggestions (they're untested though, sorry).

kitsune/src/http/handler/mastodon/api/v1/accounts/follow.rs
```diff
@@ -6,11 +6,19 @@ use crate::{
...
+    extract::{Path, Query, State},
     Json,
 };
 use kitsune_type::mastodon::relationship::Relationship;
+use serde::Deserialize;
 use speedy_uuid::Uuid;
+use utoipa::IntoParams;
+
+#[derive(Deserialize, IntoParams)]
+pub struct GetQuery {
```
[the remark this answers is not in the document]
Zeerooth: silly me, it actually was supposed to be a follow body, not query parameters, I fixed that now

kitsune/src/service/post.rs
```diff
@@ -679,27 +720,32 @@ impl PostService {
...
+                            created_at: Some(Timestamp::now_utc()),
+                        })
+                        .returning(Post::as_returning())
+                        .get_result(tx)
+                        .await?;
+
+                    Ok::<_, Error>(post)
+                }
+                .scope_boxed()
             })
```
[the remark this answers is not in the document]
Zeerooth: silly me again, there should've been code for repost notifications, I fixed that too now

Zeerooth:
@tesaguri thanks for your changes! :)

crates/kitsune-type/src/mastodon/notification.rs
```diff
@@ -0,0 +1,28 @@
+use iso8601_timestamp::Timestamp;
+use serde::{Deserialize, Serialize};
+use speedy_uuid::Uuid;
+use utoipa::ToSchema;
+
+use super::{Account, Status};
+
+#[derive(Clone, Deserialize, Serialize)]
```
aumetra, line 8:
```suggestion
#[derive(Clone, Copy, Deserialize, Serialize)]
```

crates/kitsune-type/src/mastodon/notification.rs
```diff
@@ -0,0 +1,28 @@
...
+    FollowRequest,
+    Favourite,
+    Update,
+}
+
+#[derive(Clone, Deserialize, Serialize, ToSchema)]
+pub struct Notification {
+    pub id: Uuid,
+    #[serde(rename = "type")]
+    pub _type: NotificationType,
```
aumetra, line 24:
```suggestion
    pub r#type: NotificationType,
```

And you can remove the serde rename annotation if you apply this

Zeerooth: oh nice, it's a cool trick, thanks!

aumetra:
Can you run `cargo update` once? It seems like the lockfile isn't quite up-to-date (it still references some `typed-builder` v0.15.2)

PR #333, opened 2023-09-01
Rewrite `post-processor` based on `logos`

### LINKED ISSUE #332: Rewrite post parsing
We have a somewhat suboptimal API for post parsing which requires us to use a mutex even though, logically, it shouldn't be needed.

We also use a generated pest parser. Since we don't make use of any features that go far above what a lexer can offer us, we can use something like `logos` to generate a fast lexer and then build some rudimentary facilities on top of that to allow for the transformation of the different elements.

aumetra:
Closes #332


aumetra:
Added a small benchmark to the post transformer. I don't think the transformer itself will be the bottleneck anytime soon:

```
simple_transform        time:   [284.70 ns 286.49 ns 288.15 ns]
                        change: [-0.0073% +0.6209% +1.2653%] (p = 0.06 > 0.05)
                        No change in performance detected.
```

PR #337, opened 2023-09-05
Update activity for posts

Zeerooth:
Resolves https://github.com/kitsune-soc/kitsune/issues/90

I tested federation with mastodon and it seems to be working fine.

Just like ActivityPub docs state (https://www.w3.org/TR/activitypub/#update-activity-inbox), since this should be a full replacement of the object, I allowed basically every field to be edited/reprocessed, I hope this is the desired behaviour.


Zeerooth:
Whoops, seems like I broke the check against infinite reply chains

aumetra:
Oh god, okay, we are officially running out of stack. This is bad. 

I'm gonna profile the stack sizes when I get the time and try to potentially introduce some things to take pressure off of the stack and put more on the heap.

aumetra:
Actually, I'm wondering if we could just up the stack size of the Tokio worker threads to like 4MiB (instead of keeping it at the default 2MiB)

Zeerooth:
> Actually, I'm wondering if we could just up the stack size of the Tokio worker threads to like 4MiB (instead of keeping it at the default 2MiB)

I think it might just be unavoidable, 2MB seems like not enough for any activitypub server, especially when we need to serialize and store entire threads, along with all the other overhead.

Unfortunately I probably won't have much more time this week to dedicate to kitsune, so if you can try to increase the stack size or optimize the code a bit then I'll be very grateful :> but otherwise I can tinker with it next week.

Zeerooth:
Probably the best way to reduce stack size rn would be to eliminate the recursion in `process_new_object` -> `preprocess_object` -> `fetch_object_inner` -> `process_new_object` ...

aumetra:
I'll defo commit a change for doubling the stack size of the workers on main today, the optimization is something that might take a bit longer, but definitely something that I'll look up eventually.

Especially with large futures, that can have real performance impact when a Rust move triggers a `memcpy` and it needs to copy a large future (or any large stack allocated structure)

aumetra:
Quick patch that increases the stack size to 4MiB is in #338

aumetra:
Oh damn, yeah, I forgot I just adjusted the main method.. uhm.. I guess in the case of tests it would be a linker flag.

I'm still amazed that we managed to exhaust 10MiB of stack space.

PR #343, opened 2023-09-09
Restructure to have `kitsune-core` crate

### LINKED ISSUE #312: Move services into `kitsune-core` crate
## Reasoning

We have our CLI as a dedicated binary project but we want to have things like user creation, user deletion, etc. inside the CLI, too.

Right now we would need to duplicate logic inside the CLI crate.

Another reason for moving this out is a future separate project called `kitsune-worker`. This project would be a headless job runner, enabling easy horizontal scaling by just spinning up multiple instances of that worker binary.

## Proposal

All the services and their required utility functions get moved to a new crate called `kitsune-core`.
This crate can then be used inside the `kitsune-cli` project to call things like the `UserService` or `PostService`.

aumetra:
Closes #312


kitsune/src/state.rs
```diff
@@ -109,51 +97,28 @@ impl SessionConfig {
...
-    pub event_emitter: EventEmitter,
-    pub fetcher: Fetcher,
-    #[cfg(feature = "mastodon-api")]
-    pub mastodon_mapper: crate::mapping::MastodonMapper,
-    pub oauth_endpoint: OAuthEndpoint,
-    pub service: Service,
-    pub session_config: SessionConfig,
-    pub webfinger: Webfinger,
+// BAD BAD ANTI-PATTERN BAD
+impl std::ops::Deref for AppState {
```
aumetra, line 118: Not really a "good" Rust core, to be honest. This is basically emulating inheritance

aumetra:
@Zeerooth does this look acceptable? It's a step into the direction of having a CLI with much needed functionality, such as creating users, etc. and to have a standalone job execution service

crates/kitsune-core/src/consts.rs
Zeerooth, line 1: I really feel like this file should be a part of `kitsune` as the constants there refer to api limits, we also have a figlet and env variables to get the version of the binary. It'd also be nice if the API limits could be configured, but that might be out of scope for this PR.

Zeerooth (approved):
Other than that `const` mod nit everything looks good; I'm only sad to see my boy Zustand gone :cry:

aumetra:
> I'm only sad to see my boy Zustand gone 😢

Brought 'em back 🙏

crates/kitsune-core/src/consts.rs
Zeerooth: Alright, having version and user agent in core is not a big deal, may even be beneficial as CARGO_PKG_NAME will change depending on the binary that we run so it's great if we need to include that info in logs and headers, to see where the request originates from.

> The issue here is that we are enforcing the maximum limit inside services. This gets rid of some repetition of min(requested, API_MAX_LIMIT)

Actually we still need to perform that `min` check as the builder for all these service query structs passed as a parameter usually just takes the number from the user query and doesn't check it on the spot. Maybe we could make a trait or a macro or a custom typed builder validator or sth, that just allows us to (optionally) call limit and it'll perform the `min` check, writing the lower value or the default one under the hood. But still, it'd need to be done on these service structs so we'd need to have the default limit inside the core crate. So you can keep it there for now I suppose.

aumetra: > Actually we still need to perform that min check as the builder for all these service query structs passed as a parameter usually just takes the number from the user query and doesn't check it on the spot

Yeah, that's what I meant with "performed inside the service". We are running these checks inside of the service function, such as `get_posts` where we call `.limit` with `min(get_posts.limit, API_MAX_LIMIT)`

We could continue doing the check inside the service, use a `transform` annotation via `typed-builder` or just outright *reject* invalid limits via a `garde` validator.

Using `garde` would actually enable us to:

1. Just call `.validate()` which seems clean, especially if we have multiple invariants to uphold
2. Have a simple way of having a custom API limit (we would just solve the limit by having it inside the associated `garde` context (https://docs.rs/garde/latest/garde/#contextself-access)

Zeerooth: `garde` with the context you provided really seems like a great fit for that indeed, I think we should go with it, similarly to how it's done currently with the character limit

aumetra: Okay, so I moved the default limit back into the `kitsune` crate. Gonna refactor the services to use `garde` now, and I think with that the refactor should be acceptable for now.

PR #344, opened 2023-09-10
ci: Remove `needs: fmt` requirements

tesaguri:
This allows the `fmt` job to run in parallel with the other jobs and reduces CI execution time by ~1 min.


aumetra (approved):
Thanks!

PR #347, opened 2023-09-17
Dedicated job runner

aumetra:
This PR adds a new binary that runs jobs dispatched by the main Kitsune binary.  

The idea behind this is that the running of jobs is something that can be highly parallelised, therefore someone could spin up a single server to handle all the HTTP requests, and ten job runners to handle all the outgoing federation, email delivery, etc.


aumetra:
The resulting release binary of the job runner is ~14MB. Not too bad.

PR #349, opened 2023-09-18
Resolve canonical WebFinger subjects of remote actors

### LINKED ISSUE #285: Custom WebFinger domain (à la Mastodon's `LOCAL_DOMAIN`)

tesaguri:
This patch is intended to make the ActivityPub fetcher resolve the canonical WebFinger subjects (`acct`) of remote actors and make the APIs return the resolved `acct`s so that the frontend and clients can display them.

Closes #285.

~Currently, this only updates the WebFinger resolver implementation to resolve canonical subjects and is not ready for merging yet since the ActivityPub fetcher implementation is yet to be updated. For now, the resolver is only doing more number of requests in vain.~ (Edit: Fixed in 2cea6ebe2ae3f810200a7b2ec402b819e8462a05)

That being said, I think this is at a good checkpoint for reviewing now.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3120,6 +3120,7 @@ dependencies = [
  "metrics-exporter-prometheus",
  "metrics-tracing-context",
  "metrics-util",
+ "mimalloc",
  "mime",
  "mime_guess",
  "openidconnect",
@@ -3138,7 +3139,6 @@ dependencies = [
  "strum",
  "tempfile",
  "thiserror",
- "tikv-jemallocator",
  "time",
  "tokio",
  "tokio-util",
@@ -3378,6 +3378,7 @@ dependencies = [
  "just-retry",
  "kitsune-core",
  "kitsune-db",
+ "mimalloc",
  "tokio",
  "toml 0.8.0",
  "tracing",
@@ -3459,12 +3460,12 @@ dependencies = [
  "metrics-exporter-prometheus",
  "metrics-tracing-context",
  "metrics-util",
+ "mimalloc",
  "prost",
  "rand",
  "serde",
  "tantivy",
  "tempfile",
- "tikv-jemallocator",
  "time",
  "tokio",
  "tonic",
@@ -3640,6 +3641,16 @@ name = "libm"
 version = "0.2.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f7012b1bbb0719e1097c47611d3898568c546d597c2e74d66f6087edd5233ff4"
+
+[[package]]
+name = "libmimalloc-sys"
+version = "0.1.35"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3979b5c37ece694f1f5e51e7ecc871fdb0f517ed04ee45f88d15d6d553cb9664"
+dependencies = [
+ "cc",
+ "libc",
+]
 
 [[package]]
 name = "libnghttp2-sys"
@@ -4008,6 +4019,15 @@ dependencies = [
  "migrations_internals",
  "proc-macro2",
  "quote",
+]
+
+[[package]]
+name = "mimalloc"
+version = "0.1.39"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fa01922b5ea280a911e323e4d2fd24b7fe5cc4042e0d2cda3c40775cdc4bdc9c"
+dependencies = [
+ "libmimalloc-sys",
 ]
 
 [[package]]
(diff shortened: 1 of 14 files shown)


tesaguri:
The ActivityPub actor fetcher now resolves the WebFinger `acct` name.

The patch replaces `accounts.domain` column with the resolved WebFinger domain name. Is that desirable? Or should we keep the original domain name of ActivityPub actor URI, and store the WebFinger domain in a separate column?

Also, it doesn't apply the federation filter to the WebFinger domain name yet.

aumetra:
Thanks for the PR, I'll be having a look this evening. Work has been keeping me pretty busy the last few days

tesaguri:
Another thing to consider: the new fetcher implementation falls back to `{preferredUsername}@{domain}` to keep WebFinger optional. But this could potentially lead to an inconsistent behavior if a server has multiple actors with a same `preferredUsername`, which is weird but possible.

This wouldn't be a problem if we mandate WebFinger like Mastodon does, because there is a guaranteed one-to-one correspondence between an ActivityPub actor and a WebFinger resource as a WebFinger resource only points to a single ActivityPub actor and an ActivityPub actor only points to a single `acct:` URI.

aumetra:
> The patch replaces accounts.domain column with the resolved WebFinger domain name. Is that desirable?

I think that's fine. The column is only ever used for the Webfinger-like resolver and for enforcing the "one username per instance" rule that we internally enforce to keep some consistency.

aumetra:
> Another thing to consider: the new fetcher implementation falls back to {preferredUsername}@{domain} to keep WebFinger optional. But this could potentially lead to an inconsistent behavior if a server has multiple actors with a same preferredUsername, which is weird but possible

That sounds like a very edge edgecase, no? I haven't really seen any services that don't have `preferredUsername` as a per-domain unique value.

tesaguri:
> That sounds like a very edge edgecase, no? I haven't really seen any services that don't have preferredUsername as a per-domain unique value.

Yeah it's a hypothetical talk and I think it's fine to just throw `UniqueViolation` or something in that case. Even though the spec allows it to happen, we (and many implementations out there) already reject a lot of possible-in-spec cases after all (e.g. some Activity Streams `Object` properties are considered virtually required in practice, whereas the spec [says][as-core-object] that all properties (including the `id` and `type`!) are "optional" (not the upper-case OPTIONAL though)).

The closest real-world example I know of is (now-unmaintained) Wildebeest, where usernames are treated as case-sensitive so multiple actors with a same username modulo casing can exist in a same server. But Kitsune as it stands also treats username in the same way AFAICT, and even if we were to treat *local* usernames as case-insensitive, we could still handle *remote* usernames case-sensitively, so it wouldn't be a problem in that case either.

[as-core-object]: <https://www.w3.org/TR/activitystreams-core/#object>

aumetra:
> But Kitsune as it stands also treats username in the same way AFAICT

Yeah, we do at the moment, but I feel like we should probably change that behaviour. Just way too confusing.

aumetra (approved):
Looks good! Thanks for the PR

PR #356, opened 2023-09-27
`LocalSet` optimisations

aumetra:
Optimise some parts of the code by making use of the fact that a `LocalSet` can execute `!Send` futures.

This is mostly an experiment, seeing whether the additional allocations to simply being able to send the values to the different executor thread are even worth it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -410,9 +410,9 @@ dependencies = [
 
 [[package]]
 name = "async-task"
-version = "4.4.1"
+version = "4.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


aumetra:
An example for the target of the optimisations is the parsing of the mentions. The parsing itself is very quick, microseconds even, but we need a `Mutex` to synchronize accesses to the accumulator.

That synchronization overhead could be avoided using a `LocalSet` even though the atomics overhead should be negligible.

What this would allow us is also to introduce bump allocation to certain paths in the code, since allocation arenas are inherently `!Send`.

Again, whether the overhead of the triple boxing+channel send is worth it in the end has to be checked.

PR #361, opened 2023-10-02
Add `Timestamp` swagger type

### LINKED ISSUE #360: [Swagger] Could not resolve reference: Could not resolve pointer: /components/schemas/Timestamp does not exist in document
Steps to reproduce:
1. Run kitsune
2. Go to http://localhost:5000/swagger-ui/
3. Expand the `Account` schema, for example

aumetra:
Closes #360

### DIFF
--- a/kitsune/src/http/openapi.rs
+++ b/kitsune/src/http/openapi.rs
@@ -7,7 +7,7 @@ use utoipa::{
         security::{AuthorizationCode, Flow, OAuth2, Scopes, SecurityScheme},
         OpenApi as OpenApiStruct,
     },
-    Modify, OpenApi,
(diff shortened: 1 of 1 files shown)


Zeerooth:
Huh, so it's  as easy as that

aumetra:
Yeah, `utoipa` works via namems quite a lot. So just inserting a type with the same name that resolves to a `String` works fine.

Their [`time` support](https://github.com/juhaku/utoipa/blob/4e10648a34978da243792cc2689831ace9bcec04/utoipa-gen/src/schema_type.rs#L226-L229) works be inspecting the field types via the macro and then just hardcoding `OffsetDateTime` to mean `String` in the OpenAPI schema

PR #363, opened 2023-10-03
More swagger fixes

Zeerooth:
Relates to https://github.com/kitsune-soc/kitsune/issues/360

I validated the entire schema this time and it turns out that there were still errors with `SmolStr` and `VecDeque` definitions missing + PreviewCard and PreviewType not added to the list.


aumetra (approved):
Thanks!

PR #365, opened 2023-10-07
Basic UI functionality

aumetra:
- [ ] Display single post
- [ ] Display profile
- [x] Home timeline
- [x] Local timeline
- [x] Global timeline

Note: When this is finished, we will retire our old templating-based solution for displaying profiles and posts in favour of this


aumetra:
Okay, this PR now has:

- A basic home timeline
- Basics of a post modal with WYSIWYG editing via TipTap
- A basic post view (still missing threading)
- Basic i18n in some places
- Virtually scrolling timelines
- Infinite scrolling (not yet actually fetching the new posts)

We are still missing:

- Thread fetching for the post view
- More extensive i18n
- Fetching the posts for infinite scrolling
- Profile view
- Timelines
    - Local
    - Global

aumetra:
We should also define some breakpoints as variables and use them with SCSS mixins instead of having our weird hardcoded values.

Probably best to use some widely used values such as the ones of [`Tailwind`](https://tailwindcss.com/docs/screens) or [`Bootstrap`](https://getbootstrap.com/docs/5.0/layout/breakpoints)

kitsune-fe/package.json
```diff
@@ -29,7 +33,11 @@
     "lodash": "^4.17.21",
     "pinia": "^2.1.6",
     "pinia-plugin-persistedstate": "^3.2.0",
+    "tiptap-markdown": "^0.8.2",
+    "unhead": "^1.7.4",
     "vue": "^3.2.41",
+    "vue-i18n": "^9.5.0",
+    "vue-powerglitch": "^1.0.0",
```
tesaguri, line 40: `powerglitch` doesn't seem to respect the [`prefers-reduced-motion`] media query. I doubt if the animation counts as a [motion animation] as mentioned by WCAG 2.1 Success Criterion 2.3.3 (Animation from Interactions), but I think it's worth noting anyway.

[`prefers-reduced-motion`]: <https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion>
[motion animation]: <https://www.w3.org/WAI/WCAG21/Understanding/animation-from-interactions.html#dfn-motion-animation>

aumetra: That's a good point. I think I'm gonna wrap this element inside a wrapper that implements a check for that media query

aumetra: Added the wrapper component, pretty simple one and it even reacts live on operating system preference changes

aumetra:
Merging this here. This is getting too large. Gonna split up the missing parts into smaller PRs

PR #376, opened 2023-10-16
Remove the /oauth/token endpoint from the OpenAPI scheme

Zeerooth:
Right now our schema is still invalid because of that `/oauth/token` endpoint, as [every endpoint in the OpenAPI spec needs to have at least one response defined](https://swagger.io/docs/specification/describing-responses/), which we don't do right now. One way would would be to of course write all of these possible responses and their bodies etc. but there is actually no need to. OpenAPI docs state that [including this endpoint is entirely optional](https://swagger.io/docs/specification/authentication/oauth2/#faq) so I propose just removing it from the schema. This endpoint is the same for every OAuth flow after all.

### DIFF
--- a/kitsune/src/http/openapi.rs
+++ b/kitsune/src/http/openapi.rs
@@ -77,7 +77,6 @@ struct TimestampPolyfill(String);
     modifiers(&SecurityAddon),
     paths(
         nodeinfo::two_one::get,
-        oauth::token::post,
         well_known::nodeinfo::get,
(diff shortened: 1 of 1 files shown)


Zeerooth:
I don't understand why clippy failed here lol

aumetra:
You're still importing the module. Clippy runs on deny warnings

Zeerooth:
Silly me, I did remove it that but forgot to push the commit, thanks!

PR #379, opened 2023-10-19
Compile time optimizations

aumetra:
Based on the report @Zeerooth submitted, I got rid of the newer `ring` dependency (which, for some reason, has a build script that runs for *a long time*) and got rid of the `zstd` dependency, which also takes a long time to compile native dependencies.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -270,8 +270,6 @@ dependencies = [
  "memchr",
  "pin-project-lite",
  "tokio",
- "zstd",
- "zstd-safe",
 ]
 
 [[package]]
@@ -891,7 +889,6 @@ version = "1.0.83"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f1174fb0b6ec23863f8b971027804a42614e347eafb0a95bf0b12cdae21fc4d0"
 dependencies = [
- "jobserver",
  "libc",
 ]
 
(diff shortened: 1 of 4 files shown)


Zeerooth:
Looks pretty good, a minute shaved off!

`isolang` is still the biggest offender (well, and `diesel` and `async-graphql` to a degree)

image

[cargo-timing-20231019T125507Z.html.txt](https://github.com/kitsune-soc/kitsune/files/13043055/cargo-timing-20231019T125507Z.html.txt)

aumetra:
I just rechecked on my laptop and apparently the newest `ring` has a build script that only takes 9s on my laptop.  
Therefore I'm opting for keeping `ring` at the newest version, like before.

aumetra:
(requesting one last run on your hardware, if you got the time)

Zeerooth:
@aumetra 3m 54s now

Screenshot from


[cargo-timing-20231023T083229Z.html.txt](https://github.com/kitsune-soc/kitsune/files/13067963/cargo-timing-20231023T083229Z.html.txt)

aumetra:
@Zeerooth okay, thanks for the run! I'd say that's in the acceptable range of variance.

Looking at the output, ring doesn't break the build times anymore and the additional time can be attributed to the new `opentelemetry` dependencies and to general compile time variance

Zeerooth:
Indeed, ring is not problematic anymore; out of the dependencies only isolang is the one that really takes a long time to compile, but we discussed that and there's really nothing to improve there because of these repeated long language enums

PR #394, opened 2023-10-28
fix: docker-compose start command

alexohneander:
## What this PR changes/adds
Updated `docker-compose.yaml`

## Why it does that
I have adjusted the start command so that it also executes the config that is mapped.
Additionally I have adjusted the quotation marks. If we use double quotes in the port, we should use them everywhere.


aumetra (approved):
Thanks!

PR #398, opened 2023-10-29
Make `meilisearch` enabled builds statically linked

aumetra:


aumetra:
Nice.

```
aumetra@DESKTOP-VSD7GDL:~$ ldd kitsune
        statically linked
```

tesaguri:
Generally speaking, statically linking `libssl` (or any security-sensitive library in general) into a binary distribution implies the responsibility for rebuilding the binary and cut a release when a security fix of the library is released.

In this case, though, the library is only used as a dependency of a search backend (Kitsune itself mainly uses Rustls, which shares the same characteristics as statically-linked `libssl`, but that's beyond the scope of this patch) and I don't expect the security implication would be so severe (if any). So, I'm just noting.

PR #403, opened 2023-10-29
Fix follows on Mastodon API

aumetra:


Zeerooth (approved):
aye, it makes sense to only check preferences for local accounts, thanks for fixing that!

Issue #105, opened 2023-03-14
Custom emoji support


aumetra:
General design idea:

- New configuration section called `instance`
    - This section contains `custom_emoji`, `description`, `name`, etc.

The custom emoji type could look like this:

```dhall
let CustomEmoji = { name: Text, path: Text }
let CustomEmojis = List CustomEmoji
in {
     CustomEmoji, 
     CustomEmojis
}
``` 

The custom emoji would then be mounted in some media sub-path, something like `/media/custom-emoji/:emojiName`, and would then, as shown in the path, indexed by their name  
Not sure if this might be something that'd be better kept in the database, since this might need to synchronise between multiple nodes.

This is just a first design draft, completely open for discussion

aumetra:
Note that this also needs database schema changes and new ActivityPub types

aumetra:
Support to load emoji used inside an object was added #110

Zeerooth:
Would remote emojis be something that we'd like to support? That's very much feasible to do and IMO it'd be quite a cool and useful feature, seeing how often emoji packs are duplicated between instances and how sometimes missing out on emojis not present on your current instance sucks. 

The downside would be that potentially malicious actors could pull offensive emojis from some other server and disrupt the local instance, so it's necessary to add endpoints moderating the federated emojis + configuration variable for blocking that feature.

Mastodon issue reference: https://github.com/mastodon/mastodon/issues/8649

aumetra:
Remote emoji would be interesting but it'd be a challenge of how to actually do the discovery of these emoji.

Do we store them when receiving them? Do we allow a mention syntax (ala `:blob-dj@my-instance.example:`)?

In case we use this mention syntax, how do we discover them? Do we just assume each remote server with custom emoji implements the Mastodon API? What about Misskey/IceShrimp?

aumetra:
Another much more basic question, how do we configure these custom emoji? Configuration file? Datababse?

Zeerooth:
> Another much more basic question, how do we configure these custom emoji? Configuration file? Datababse?

I was thinking of doing it through database. Have a new table for storing emojis that refers to media_attachments and a new service for adding, modifying and removing emojis. Eventually when we have a new kitsune-cli we can provide a way to add emojis through cli or maybe have a graphql endpoint and then hook it up to UI in the future. 

I don't think using a configuration file for that is a good idea because we'd need to restart kitsune just to modify emojis which is really not needed.

[...]

PR #405, opened 2023-10-29
Custom and remote emojis 

Zeerooth:
It's not ready to be merged or even reviewed, but I'd like to already open this draft PR to keep track of the progress and address potential feedback. 

How it works:
- Server saves all the emojis that it gets through through federated posts (to confirm that they are genuine they get resolved as an AP type, just like accounts are)
- Users then are able to use :emoji@example.com: syntax to insert the emojis into posts (unfortunately it doesn't work for emojis that the server doesn't know about yet as there's no webfinger like for the accounts)
- These emojis get translated into :emoji__example_com: (neither mastodon nor misskey allow special characters like @ or . which sucks) and inserted into the tag list, with id and image url pointing to the original resource.
- Voila, it works

Things that are already implemented and work:
- [x] Parser for emojis in posts
- [x] DB tables 
- [x] Barebones emoji service

Things that probably work but need to be tested:
- [x] Emoji federation - mastodon actually exposes emojis as a resolvable AP type. I implemented a basic resolver ~~but haven't tested it yet. Also who knows how it behaves with misskey or pleroma for example.~~ it works!

To be added:
- [x] GET /api/v1/custom_emojis
- [x] GET /api/v1/statuses/{id} <- needs the emoji list
- [x] Update emojis during post updates 
- [x] More tests
- [x] Expose the emojis as a resolvable AP type

Out of scope for now (I'll make an issue for it later)
- Resolve emojis in account properties, such as description, display name, fields
- Make it easy to add local emojis through the CLI and/or admin panel (if we ever end up making one)

Issue ref: https://github.com/kitsune-soc/kitsune/issues/105


aumetra:
It indeed does work, heh

Zeerooth:
hmmm I have no idea why the tests keep timing out

Zeerooth:
@aumetra could you take a look at this PR when you've got some time pls?

aumetra:
> could you take a look at this PR when you've got some time pls?

Will do on my commute home

crates/kitsune-core/src/activitypub/mod.rs
```diff
@@ -63,6 +67,48 @@ async fn handle_mentions(
...
+            .as_ref()
+            .map(|remote_id| fetcher.fetch_emoji(remote_id))
+    });
+
+    let emojis = join_all(futures)
+        .await
+        .into_iter()
+        .collect::<Result<Vec<_>, _>>()?
+        .iter()
+        .zip(emoji_iter)
```
aumetra, line 95: This `.zip` here doesn't seem quite correct. If emoji were filtered in the `.filter_map` step above, there would be a mismatch between the left and right values.

Consider passing them down inside a tuple from the future.

crates/kitsune-core/src/mapping/activitypub/object.rs
```diff
@@ -224,3 +266,40 @@ impl IntoObject for Account {
...
+
+#[async_trait]
+impl IntoObject for CustomEmoji {
+    type Output = Emoji;
+
+    async fn into_object(self, state: &State) -> Result<Self::Output> {
+        // Officially we don't have any info about remote emojis as we're not the origin
+        // Let's pretend we're not home and do not answer
+        let name = match self.domain {
+            None => Ok::<String, ApiError>(format!(":{}:", self.shortcode)),
```
aumetra, line 278: Small sidenote: Do we need that type annotation here?

crates/kitsune-core/src/mapping/activitypub/object.rs
```diff
@@ -224,3 +266,40 @@ impl IntoObject for Account {
...
+            .db_pool
+            .with_connection(|db_conn| {
+                async move {
+                    media_attachments::table
+                        .find(self.media_attachment_id)
+                        .get_result::<DbMediaAttachment>(db_conn)
+                        .map_err(Error::from)
+                        .and_then(|media_attachment| media_attachment.into_object(state))
+                        .await
+                }
```
aumetra, line 291: Async block indirection. We don't need that here

crates/kitsune-core/src/mapping/mastodon/sealed.rs
```diff
@@ -589,3 +613,37 @@ impl IntoMastodon for DbNotification {
...
+
+#[async_trait]
+impl IntoMastodon for (DbCustomEmoji, DbMediaAttachment, Option<Timestamp>) {
+    type Output = CustomEmoji;
+
+    fn id(&self) -> Option<Uuid> {
+        Some(self.0.id)
+    }
+
+    async fn into_mastodon(self, state: MapperState<'_>) -> Result<Self::Output> {
```
aumetra, line 625: Consider destructuring `self` inside the block for more readable code.

I.e `let (emoji, attachment, timestamp) = self;`

crates/kitsune-core/src/service/custom_emoji.rs
```diff
@@ -0,0 +1,190 @@
...
+                    .or(posts::created_at.is_not_null()),
+            )
+            .distinct_on(custom_emojis::id)
+            .select((
+                CustomEmoji::as_select(),
+                MediaAttachment::as_select(),
+                posts::created_at.nullable(),
+            ))
+            .limit(get_emoji_list.limit)
+            .into_boxed();
```
aumetra, line 141: Any reason we box here?

lib/post-process/src/lib.rs
```diff
@@ -41,6 +41,24 @@ fn enforce_prefix<'a>(lexer: &Lexer<'a, PostElement<'a>>) -> bool {
     }
 }
 
+#[inline]
+fn emoji_split<'a>(lexer: &Lexer<'a, PostElement<'a>>) -> Option<(&'a str, Option<&'a str>)> {
+    if !enforce_prefix(lexer) || !enforce_postfix(lexer) {
```
aumetra, line 46: Actually we don't need to enforce prefixes and postfixes here, since custom emojis are valid, even when not separated properly from other post components

aumetra (commented):
Some initial comments. I'll look at it closer later

crates/kitsune-core/src/activitypub/mod.rs
aumetra: That should also make it possible to get rid of the intermediate collect

Zeerooth:
@aumetra thanks so much for these comments, I addressed everything now :)

BTW there is one more aspect of this PR that I was thinking about - for the custom emoji table I decided to not save the domain for local emojis (it's NULL), but I saw that in other tables, for example for user accounts, the domain part is not nullable and saved even for local accounts. Should I change it accordingly for custom emojis?

aumetra:
> Should I change it accordingly for custom emojis?

I don't think that's necessary here. We store the domains on accounts and users because of potential multi-tendant support (i.e. one instance with multiple domains).

But not everyone will be allowed to add custom emoji anyway, just the moderators and admins of the overall system, so might as well not store the domain here.

Zeerooth:
Right, then I'll leave it as it is

aumetra:
@Zeerooth would you mind rebasing this? I don't really have any more comments to this

aumetra (approved):

aumetra:
Merged main into it. Do you have any last things you wanna add before I hit the merge button?

Zeerooth:
> Merged main into it. Do you have any last things you wanna add before I hit the merge button?

Not really, just that you're too quick, I merged it myself but then I try to push and see that you already pushed to this branch lol

Thanks though, you can merge it now :)

PR #407, opened 2023-10-29
Make account names case-insensitive

### LINKED ISSUE #406: Make usernames case insensitive

aumetra:
Closes #406 

---

- [x] Names are insensitive in the `accounts` table
- [x] Names are insensitive in the `users` table
- [x] ~~Convert the fields to `citext` (https://www.postgresql.org/docs/current/citext.html)~~
- [x] Use Unicode Collation <https://www.postgresql.org/docs/current/collation.html>
- [x] Adjust regex in the registration function to use `\p{L}\p{N}` instead of `\w`
- [x] Tests added
    - [x] `accounts`
    - [x] `users`
    - [x] Registration garde validator


aumetra:
Better yet: instead of `CITEXT`, we can use Unicode collations. That works with things other than ASCII 

https://www.postgresql.org/docs/current/collation.html

crates/kitsune-db/migrations/2023-05-19-192931_initial_tables/up.sql
```diff
@@ -33,7 +33,7 @@ CREATE TABLE accounts (
     ) STORED NOT NULL,
 
     -- UNIQUE constraints
-    UNIQUE (username, domain)
+    UNIQUE (LOWER(username), domain)
```
tesaguri, line 36: IIUC the case-insensitive unique constraint applies to remote accounts as well. Is that intentional?

Among existing major implementations that I know, only Wildebeest allows case-insensitively equal usernames to coexist in a server, and it seems to be unlikely that any serious future implementation would behave the same way, so I think it might not be too bad, if intentional.

---

Also, the current implementation stores remote domain names as resolved via WebFinger, and we might want to consider their case-sensitivity as well.

aumetra: Yeah, that was intentional. And good catch regarding the domain.
I'm right now thinking about just lowercasing the domain when receiving.

Domains are case insensitive anyway.

For usernames I actually want to start using non-deterministic Unicode collations. Because that way `🅰️umetra`, `aumetra`, and `AumEtra` are considered the same name.

Maybe I'm gonna add insensitivity for accent characters as well, to avoid the ambiguity between ä and a, for example.

Because I'd kinda like to keep the ability to allow for non-ASCII usernames. I don't know, just seems fun.

tesaguri: > Because I'd kinda like to keep the ability to allow for non-ASCII usernames. I don't know, just seems fun.

IIRC Weibo allows Chinese characters for their counterpart of usernames, so I wouldn't be surprised if there appears a Weibo-clone from the Chinese-speaking world that extensively uses non-ASCII usernames.

aumetra: Also, regarding the domain, do you think a punycode normalization would be useful?

I thought about maybe just shoving every domain through the `idna` crate since that gives us either the Unicode or the ASCII representation. 

One normalized form. And the crate also adjusts uppercase, and other stuff

tesaguri: I'm not sure to be honest, but since remote WebFinger endpoints in the wild might return any form of domain names (whether that's compliant to standards or not), it may be the safest to normalize everything.

aumetra:
Looking at the current poll results I put out on Mastodon, we will probably keep an alphanumeric check in place:

https://corteximplant.com/@0x0/111325717243948397

image

(yes, the results are saying "any unicode character" but the issue here is, as @tesaguri [pointed out](https://fedibird.com/@tesaguri/111327462805995336), the compatibility with the `acct:` URI scheme)

PR #413, opened 2023-11-08
Use nightly features

aumetra:


aumetra:
Locally this thing doesn't stack overflow 🤷‍♀️

aumetra:
Anyway, time to whip out `stacker` and `println` debug

PR #421, opened 2023-11-11
Multiple federation protocols: Code restructure

aumetra:
Restructure the code towards the support for multiple federation protocols


aumetra:
So the plan for `kitsune-core` is to literally rip most of the components out of it.  
Services go into an own crate, resolvers go into own crates, etc.

The core crate will pretty much only be there to define traits for the rest of the components, to provide generic interop.

aumetra:
The `kitsune-activitypub` crate shouldn't directly reference `kitsune-webfinger`, but instead something generic that satisfies the `Resolver` trait, but we aren't there yet where this trait is defined.

aumetra:
Ugh.. now comes the awkward struggle where I have to somehow share parts of the services crate with the jobs crate in an attempt to use the mailing and user services..

PR #440, opened 2023-12-07
CSRF protection on OAuth pages

### LINKED ISSUE #431: Add CSRF token to OAuth authorization form

aumetra:
Closes #431 

This PR adds:

- [x] A generic CSRF protection library (`lib/cursiv`)
- [x] Usage of the CSRF token on the OAuth pages


aumetra:
Thinking of using a keyed BLAKE2/BLAKE3 hash here. Fast and simple.  

BLAKE3 is superior in terms of raw speed (according to benchmarks and (IIRC) available implementations in Rust), but I don't think that the performance difference will really show in this case, since the values we work on are small.

PR #455, opened 2023-12-16
Hyper v1 upgrade

### LINKED ISSUE #436: Path to hyper v1
Blockers:

- [ ] `hyper-rustls`: https://github.com/rustls/hyper-rustls/issues/234
- [x] `async-graphql`: https://github.com/async-graphql/async-graphql/issues/1430
- [ ] `openidconnect`
- [ ] `opentelemetry-http`
- [x] `oxide-auth`

aumetra:
Closes #436


crates/kitsune-http-client/src/lib.rs
```diff
@@ -251,7 +237,15 @@ impl Client {
...
     /// - The inner client service isn't ready
     /// - The request failed
-    pub async fn execute(&self, req: Request<Body>) -> Result<Response> {
+    pub async fn execute<B>(&self, req: Request<B>) -> Result<Response>
+    where
+        B: HttpBody<Data = Bytes> + Send + Sync + 'static,
+        B::Error: Into<BoxError> + 'static,
+    {
+        let (parts, body) = req.into_parts();
+        let body = BoxBody::new(body.map_err(Into::into));
```
tesaguri, line 246: I believe the type-erasure can be done in a separate function to reduce the monomorphisation cost.

crates/kitsune-http-client/src/lib.rs
```diff
@@ -270,14 +264,20 @@ impl Client {
...
+        req: Request<B>,
         private_key: PrivateKey<'_, K>,
     ) -> Result<Response>
     where
+        B: HttpBody<Data = Bytes> + Send + Sync + 'static,
+        B::Error: Into<BoxError> + 'static,
         K: SigningKey + Send + 'static,
     {
+        let (parts, body) = req.into_parts();
+        let body = BoxBody::new(body.map_err(Into::into));
```
tesaguri, line 278: Same here.

crates/kitsune-webfinger/tests/redirects.rs
```diff
@@ -64,12 +65,12 @@ async fn reject_fake_jrd_redirect() {
...
                 })
                 .unwrap();
-                Ok::<_, Infallible>(Response::new(Body::from(body)))
+                Ok::<_, Infallible>(Response::new(BoxBody::new(Full::new(Bytes::from(body)))))
             }
             ("whitehouse.gov", "/.well-known/webfinger?resource=acct:0x0@whitehouse.gov") => {
                 Ok(Response::builder()
                     .status(StatusCode::NOT_FOUND)
-                    .body(Body::empty())
+                    .body(BoxBody::default())
```
tesaguri, line 73: You can use the `Default` impl of `Full` instead, which creates an empty body.

aumetra: Applied

aumetra:
Built around the missing libraries by writing the ugly temporary `http-compat` crate.  
Works well enough. Not my favourite escape hatch, but there's nothing else we can do right now (except waiting which I don't want to anymore).

aumetra:
Something I still want to take care of is to take a look at the ported code of the `kitsune-http-client` crate.  
I feel like there is something we can improve here. But yeah, for the most part, I'd say this is the final version.

aumetra:
Would be great if you could take a little skim over it :D  
No pressure tho. It's not urgent

tesaguri (changes requested):
Looks good to me, overall!

However, I believe the soundness nit needs some attention.

crates/kitsune-http-client/src/lib.rs
```diff
@@ -428,8 +439,15 @@ impl Response {
...
     pub fn stream(self) -> impl Stream<Item = Result<Bytes>> {
-        BodyStream {
-            inner: self.inner.into_body(),
+        try_stream! {
+            let body_stream = BodyStream::new(self.inner.into_body());
+
+            for await frame in body_stream {
+                match frame.map_err(Error::new)?.into_data() {
+                    Ok(val) if val.has_remaining() => yield val,
+                    _ => (),
```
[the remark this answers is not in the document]
aumetra: Clarified this with an explicit match and a comment.

crates/kitsune-http-client/src/lib.rs
```diff
@@ -162,13 +150,19 @@ impl ClientBuilder {
...
 
-        let client: HyperClient<HttpsConnector<HttpConnector>, Body> =
-            HyperClient::builder().build(connector);
+        let client = HyperClient::builder(TokioExecutor::new())
+            .build(connector)
+            .map_response(|resp| {
+                let (parts, body) = resp.into_parts();
+                let body = BoxBody::new(body);
+                HyperResponse::from_parts(parts, body)
+            });
```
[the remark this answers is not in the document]
aumetra: Yeah, I thought about writing an own little body type that wraps multiple types under the hood.  
Something like this

```rust
enum Body {
    Empty,
    Full(Bytes),
    Stream(StreamBody<Bytes, BoxError>),
}
```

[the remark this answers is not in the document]
aumetra: I think @novacrazy from Lantern chat had a similar issue when upgrading to v1. Just the general annoyance that the general `hyper::Body` type is now gone.

tesaguri (approved):

aumetra:
Actually, I think I'm gonna make a little own body type. Something like the old `hyper::Body` on a budget.  
Should simplify some code and make for a nicer DX when using the client.

tesaguri:
> Opinion on this new body type? Better or worse than just using generic type parameters?

If we want to have that body type for ergonomics, I expect we would need to write some boilerplate conversion trait implementations, and I'm not sure if the ergonomics gain would outweigh the implementation cost until actually write one.

Personally, I'd prefer to try to contribute such a general-purpose code to the upstream framework first. There was an idea of a `GenericBody` type during the discussion on the split of the now-defunct `hyper::Body` type (<https://github.com/hyperium/hyper/issues/2345#issuecomment-734982106>), which hasn't been realized (yet). Perhaps that's just because nobody hasn't worked on it yet, and if so, I guess there's a chance of a PR getting accepted. But that would take a considerable amount of time even if ever accepted, and might be too much yak-shaving ツ.

aumetra:
@tesaguri well, I wrote one here https://github.com/kitsune-soc/kitsune/commit/c00b91e2595591c6f86d8b2ea2c9e64c485032fd
What's your opinion on that one? It's kinda similar-ish to what we had with `hyper::Body`

tesaguri (approved):
Looks good to me!

aumetra:
Perfect! Merging this then

Issue #418, opened 2023-11-11
Full restructure


aumetra:
So, Kitsune right now is pretty intertwined with ActivityPub and Webfinger. Kinda makes sense since it started out by being built around those standards, but I think we can do better with minimal performance impact.

## Rough plan

1. Move ActivityPub logic into `kitsune-activitypub`
2. Define traits in `kitsune-core`, such as the `Federator`, `FederationFetcher`, and `Resolver` trait
    - These three areas were chosen since they define pretty well how ActivityPub works:
        1. Resolve an actor via Webfinger
        2. Fetch them via a fetcher
        3. Use the federator to send out things such as post creation, deletion, favourites, etc.
3. Instead of giving services a `Fetcher` or `Deliverer`, we give them a vector of these which they iterate over and perform the appropriate actions in the different federated networks
4. Restructure the database
    1. Create a separate key table to store the keys and their IDs and references to an account that owns them
    2. Move the ActivityPub-specific things, such as collection URLs into its seperate table
    3. Rename the `actor_type` field to something more generic (such as just `type`)
    4. Add a `protocol` marker to each account, post, like, etc. to identify from what network they originate from

---

I invite some bikeshedding to this issue since this is a *large* undertaking and will make *significant* changes to all parts of Kitsune

aumetra:
This we still need to address here are system actors (cc @perillamint, if you have any suggestions)

aumetra:
For multiple domains, we could have a table such as this one:

```sql
CREATE TABLE domains (
    domain TEXT PRIMARY KEY,
    owner_id UUID REFERENCES users(id),
    globally_usable BOOLEAN NOT NULL, -- Name can be bikeshed. Marker on whether anyone from the instance can choose this as "their" domain
    verification_token TEXT NOT NULL,
    verified_at TIMESTAMPTZ
);

aumetra:
(gonna try to land #405 before starting this because rebasing on these planned changes would be a *pain*)

Zeerooth:
If we're doing this then I think it's important to think about how to potentially tie https://github.com/kitsune-soc/kitsune/issues/134 into our future account schema

aumetra:
> it's important to think about how to potentially tie https://github.com/kitsune-soc/kitsune/issues/134 into our future account schema

Do you have any suggestions on how we can tie this in?  
Potentially a new table, such as `account_proofs` with the structure of:

```sql
CREATE TABLE accounts_proofs (
[...]

Issue #190, opened 2023-04-21
Nomadic Identity


erlend-sh:
https://socialhub.activitypub.rocks/t/defining-alsoknownas/907
https://socialhub.activitypub.rocks/t/nomadic-identity-for-the-fediverse/2101/
https://codeberg.org/fediverse/fep/src/branch/main/feps/fep-c390.md

Example implementation of fep-c390: https://codeberg.org/silverpill/mitra

Implementing this fep for Kitsune to be compatible with Mitra would set a strong precedent for a stable fep proposal, ready for wider adoption.

aumetra:
Related to #134

erlend-sh:
Ah there it is. I had forgotten about the ‘identity proofs’ moniker when I looked ahead of posting. Feel free to just merge these links in there and delete this issue, as it may be redundant.

erlend-sh:
https://socialhub.activitypub.rocks/t/fep-7952-roadmap-for-actor-and-object-portability/4332

[https://codeberg.org/bumblefudge/fep/src/branch/fep-7952--roadmap-for-actor-and-object-portability/fep/7952/fep-7952.md](Proposal text)

For anyone else like me who struggled to find the full proposal text, here it is:

https://codeberg.org/bumblefudge/fep/src/branch/fep-7952--roadmap-for-actor-and-object-portability/fep/7952/fep-7952.md

And the prime objective of the FEP is summed up pretty well here:

> ### Unbundle the services and concerns of a typical instance
> 
> 1. **Sign everything**: Recognize client-side cryptographic signatures as proof of authorship (by implementing [FEP-8b32: Object Integrity Proofs](https://codeberg.org/fediverse/fep/src/branch/main/fep/8b32/fep-8b32.md)), in addition to the current practice of relying solely on the instance URL.
> 2. **B.Y.O. Actor ID**: Using Object Integrity proofs enables Identity Hosting to be separated from the other instance concerns. Actor profiles can now be hosted separately from the instance (including as a static JSON object on a personal website), which in turn enables service providers to offer their users a “BYO (Bring Your Own) domain name” feature.
> 3. **Separate Inbox/Outbox**: (Optional) The previous steps enable message transfer and Inbox/Outbox hosting to be outsourced to separate service providers (the Actor profile links to these in the usual manner).
[...]

PR #458, opened 2023-12-16
Database restructure

aumetra:
Part of #418 

Rough outline:

- [x] Create a separate key table to store the keys and their IDs and references to an account that owns them
- [x] Move the ActivityPub-specific things, such as collection URLs into its seperate table
- [x] Rename the actor_type field to something more generic (such as just type; renamed to `account_type`)
- [x] Remove `blurhash` column (not sure if we even want blurhashes)
- [x] Move domains to a separate table, designate owners, and whether they can be globally used
- [x] Revisit the roles table and make it suitable for custom roles 
- [x] Add a protocol marker to each account ~~, post, like, etc.~~ to identify from what network they originate from (rationale: knowing the account is enough to know how to federate)

TODO:

- [ ] Review the cryptographic keys name and reference structure
- [x] Get third-party review


aumetra:
Structure is looking way too complicated now.. kinda..
kitsune_dev

aumetra:
Do we need a protocol marker on each element? Maybe not. I think having it for the account itself is sufficient. It doesn't matter where the elements in particular come from, does it?

aumetra:
Ah, yeah, for the domain stuff.. I need to vendor my DNS verification library into the monorepo at some point..

erlend-sh:
For the two remaining TODOs..
> * Review the cryptographic keys name and reference structure
> * Get third-party review

..couldn't this wait until *after* the initial merge? Kitsune isn't used *for real* in production yet anyhow. It'll be a lot easier to get eyes on this and stir up interest among prospective reviewers if the WIP is merged and put to the test.

aumetra:
Oh yeah, probably can. I sent the diagram to @perillamint already, they are looking at it next time they have time. After that I'll just hit merge on it and see what we do after that.

aumetra:
The "third party review" is mostly about "getting someone else to look at the database schema that I didn't overlook anything or completely overcomplicated everything

perillamint:
Some limitations I noticed:

* This structure can't hold language code that does not adhere to ISO-639 (because it uses enum)

Also, I think it would be good to have `name` and `public_visibility` column in `user_roles` table (to display role badges on user profiles)

Additionally, it seems the current schema does not have a way to store pinned user notes.

For Mastodon API compatibility, I think it needs `posts_bookmark` to support the bookmark feature

aumetra:
> This structure can't hold language code that does not adhere to ISO-639 (because it uses enum)

I think this is a fair tradeoff regarding storage and computing vs. storing non-normalized data

> Also, I think it would be good to have name and public_visibility column in user_roles table (to display role badges on user profiles)

We have the `name` of the role inside the `roles` table. The `users_roles` table is just a PK-FK mapping table to allow for M:N mapping of users <-> roles.  

Not sure if we need the visibility in this iteration. We can always add them later and then collapse the migrations into one each release, basically saying "you need to update in lock-step from version to version" (similar to what Nextcloud does with its major versions, where you can only update one at a time)

perillamint:
> > This structure can't hold language code that does not adhere to ISO-639 (because it uses enum)
> 
> I think this is a fair tradeoff regarding storage and computing vs. storing non-normalized data

I think if it is nullable, I don't think it wouldn't be problem (to handle Misskey posts (they don't send notes with locale property) and weird locales)


> > Also, I think it would be good to have name and public_visibility column in user_roles table (to display role badges on user profiles)
> 
> We have the `name` of the role inside the `roles` table. The `users_roles` table is just a PK-FK mapping table to allow for M:N mapping of users <-> roles.
> 
> Not sure if we need the visibility in this iteration. We can always add them later and then collapse the migrations into one each release, basically saying "you need to update in lock-step from version to version" (similar to what Nextcloud does with its major versions, where you can only update one at a time).

I think it could be added after we decide to implement role visibility.

Additional information about its use cases: I saw some cases when some fedmins mark admin account with label (and show it publically to mark it as official account and handle inquiries) or some Misskey admins are giving their users with some kind of "joke" roles

> But about the bookmark table, that's fair. That's something I'll add now. Because it's an endpoint I want to support in the near future (and this time actually persistent bookmarks, not like the mess Mastodon has where bookmarks just disappear).

:+1:

aumetra:
> I think if it is nullable, I don't think it wouldn't be problem (to handle Misskey posts (they don't send notes with locale property) and weird locales

Actually, my idea here was to try to guess the locale using the language detection backend we have (so either via [`whatlang`](https://docs.rs/whatlang/latest/whatlang/) or [`whichlang`](https://docs.rs/whichlang/latest/whichlang/)).

That way we can have somewhat accurate predictions and default to English otherwise. Not sure if we should default to English or introduce nullability..

perillamint:
Btw, one question: Since we have to sanitize HTML from the ActivityPub status, I think it would be a good idea to store sanitized HTML in separated form (because running ammonia (or other sanitizer) every time on client request would be quite expensive).

Is `posts.content` and `posts.content_source` meant for that purpose?

crates/kitsune-db/migrations/2023-12-16-192941_initial/up.sql
```diff
@@ -0,0 +1,771 @@
...
+
+CREATE TABLE posts
+(
+    id               UUID PRIMARY KEY,
+    account_id       UUID                                                                                                   NOT NULL,
+
+    in_reply_to_id   UUID,
+    reposted_post_id UUID,
+
+    is_sensitive     BOOLEAN                                                                                                NOT NULL,
```
tesaguri, line 334: Misskey sets the `sensitive` flags on a per-medium basis, i.e. authors can attach non-sensitive media and sensitive media to a single post at the same time. Is this a use case Kitsune would like to support?

tesaguri:
#190 requires the original activity JSONs (for signed local posts at least) to be stored in the database in order to serve client-signed objects, IIUC.

crates/kitsune-db/migrations/2023-12-16-192941_initial/up.sql
aumetra: Good idea actually.. and for Mastodon, where we have to have a single sensitive flag, we simply do an OR chain for the `is_sensitive` (or an `.any` on an iterator, in Rust terms :P)

aumetra:
The preferences table, which is really just a bunch of booleans, can probably be refactored into something like a bitflag thing.

```rust
bitflags! {
    pub struct Notify: i64 {
        const ON_FOLLOW = 0b00000001;
        const ON_MENTION = 0b00000010;
        const ON_LIKE = 0b00000100;
    }
}

pub struct AccountsPreference {
    pub account_id: Uuid,
    pub notify: Notify,
}
```

Saves space. Makes stuff more readable. Gets rid of a lot of boolean comparison stuff.

aumetra:
Scratch the bitflag idea. Can't index it, can't efficiently search it. Gonna stick to bools. I guess it doesn't even save that much storage since the database can do the efficiency improvements.

Issue #477, opened 2024-01-26
[Tracking] Relicensing


aumetra:
## Rough outline

- Repository in general will be relicensed under `AGPL-3.0`
- Code in the `lib/` directory will be licensed under `MIT OR Apache-2.0` for compatibility with the wider Rust ecosystem

## Open questions

- [x] Should the license be `AGPL-3.0-only` or `AGPL-3.0-or-later`? (decision: we are using `AGPL-3.0-or-later`)
- [x] Should we use a per-file licensing scheme? (preliminary decision: per-crate/per-directory licensing scheme)

## Potentially useful tools

- [FSFE REUSE tool](https://github.com/fsfe/reuse-tool)

---

The general idea behind this tracking issue is to get some of the "private" Matrix discussions more out into the public and to keep them more persistent.

---

cc @erlend-sh @tesaguri @perillamint

aumetra:
@perillamint You mentioned per-file licensing. How easy is it to keep compliance with a setup like that?  
I imagine it being a little more complicated, no? Because it's easier to accidentally link against an AGPL licensed component from an MIT file.

But we can use the REUSE tool you proposed to add SPDX identifiers to each file to make it easier to see at a glance what license that file has.  
The only issue I have with the tool is that it adds the copyright holder and year. Especially the year seems problematic since every year you would have a gigantic commit touching every file in the repository.

PR #478, opened 2024-01-27
Move `kitsune-retry-policies` into `lib/` directory

aumetra:
Preparation for the relicensing (see #477)

This moves the `kitsune-retry-policies` crate into the `lib/` directory since it is used by a library (`athena`) which would be licensed under `MIT-OR-Apache-2.0` under the new licensing scheme, but the `crates/` directory would be licensed under `AGPL-3.0`, making them incompatible.


aumetra:
Bikeshedding the name though. The crate isn't *inherently* tied to Kitsune, just some opinionated defaults we use.

aumetra:
For now the name doesn't matter. License compliance is a bit more important

PR #481, opened 2024-01-28
Make language detection more configurable

### LINKED ISSUE #480: Extend configurability of the language detection backends
**Is your feature request related to a problem? Please describe.**

Kitsune has support for multiple language detection backends (currently `whatlang` and `whichlang`) but they are mostly configurable via compile-time flags.  
This is annoying for people because:

1. They probably don't wanna fully recompile when they want to switch between backends
2. This doesn't work for people who use prebuilt binaries
3. Some people just don't want to get into the technicalities on how to recompile Rust binaries with specific feature flags

**Describe the solution you'd like**

1. Compile against both, `whichlang` and `whatlang`
2. Expose another configuration section called `language-detection`
3. Make the backend configurable via a stringly-typed TOML enum

Another nice feature flag would be the language it defaults to. At the moment the language (if it wasn't possible to reliably determine it) falls back to the default value of english.  
While this works great for a large chunk of people (namely, all of the english speaking population), it doesn't work as well for instances that have a different primary language.

This would allow mainly Korean instances to set the fallback to Korean, German instances to German, Japanese instances to Japanese, etc.

aumetra:
Closes #480 

---

Not very beautiful. I don't like how much I have to pass this configuration struct around.


aumetra:
Oh, cool. Nice edge-case we discovered here. `whichlang` apparently panicks when it receives an empty string. Very cool.

PR #485, opened 2024-02-09
Custom Redis pool

### LINKED ISSUE #482: Switch from `deadpool-redis` to a home-grown alternative
The next version of `redis-rs` will deprecate `aio::Connection` in favour of its multiplexed counterpart.  
With that deprecation there is *no* benefit of using a traditional pooling system like `deadpool` since these systems are built on the premise of quickly "borrowing" something from a pool and then returning the object.

These borrows are exclusive since they give you temporary ownership, which isn't utilizing a multiplex-ready connection optimally.

---

## Solution

A new custom solution which internally looks something like this:

```rust
trait Strategy<Obj>: Send + Sync {
    fn choose(&self, objs: &[Obj]) -> &Obj;
}

struct PoolInner<Obj> {
    strategy: Box<dyn Strategy<Obj>>,
    objects: Box<[Obj]>,
}

pub struct Pool<Obj> {
    inner: Arc<PoolInner<Obj>>,
}

impl<Obj> Pool<Obj> {
    pub fn acquire(&self) -> Obj {
        let chosen = self.inner.strategy.choose(&self.inner.objects);
        chosen.clone()
    }
}
```

Something like that. This will use a strategy to choose an object from a slice of objects which is then subsequently cloned.  
An example strategy could be a round-robin strategy using an atomic integer to increment modulo the pool length so you don't have to reset the counter ever.

That way you can ensure an optimal utilization of all connections since in a highly contented scenario, connections can serve N in-flight requests at the same time.

aumetra:
This PR introduces a new small crate that manages a batch of connections.  
These connections have to manage their health and reconnections themselves which the `redis` crate fortunately offers via the `ConnectionManager` type.

This type reconnects automatically on connection failure and multiplexes. Therefore we can simply clone this type and hand it out to consumers and it will just work.

It should also improve the utilization of connections through multiplexing, making Kitsune make do with less connections than before.

---

Closes #482

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -246,7 +246,7 @@ dependencies = [
  "askama_derive",
  "askama_escape",
  "humansize",
- "num-traits 0.2.17",
+ "num-traits 0.2.18",
  "percent-encoding",
 ]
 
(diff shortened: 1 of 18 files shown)


aumetra:
Actually.. wouldn't it be better to use `bb8-redis` with the multiplex manager since it only has one manager for these connections instead of N separate managers?

Ugh.. I don't know..

aumetra:
@perillamint @tesaguri any opinions on that?

PR #488, opened 2024-02-11
HTTP signatures rewrite

### LINKED ISSUE #195: HTTP signature hardening
Continuation of #194 

- [x] Required headers (for example, `POST` requests require `Digest`, `Date`, `(request-target)`)
- [x] Maximum signature age
- [ ] Extended owner checks 
      (Activity owner == Object owner if object is embedded; reject if object is not embedded for certain activity types)

aumetra:
Rewrite of the HTTP signatures implementation using a more sophisticated parsing method and implementing hardened rules to hopefully get rid of any security issues that could arise from improperly implemented HTTP signatures.

Closes #195


aumetra:
# Design

This crate will expose two different API levels. A mid-level API and a high-level API.  

## Mid-level API

This API provides header parsing, signing string construction and verification/signing, etc. but all in separate steps.

Mostly geared towards either:

1. People who only need a part of the entire validation/signing pipeline
2. People who can't use use high-level API since they have lifetime issues, aren't using async, etc.

## High-level API

The high-level API will expose a conservative set of configuration options and will have an API akin to this:

```rust
async fn (req, opts) -> bool
```

where the boolean implies whether the verification succeeded or failed (for whatever reason).  
It exposes no metadata, no failure reasons, etc. in order to keep any crypto failures due to leakage of metadata from happening.

It's an async method and will execute the actual validation on a rayon background task. Key fetching is only done when required using a closure which returns a fallible future.  
That way it will integrate neatly with all kinds of asynchronous operations you'd do on a webserver to fetch a key (be it a database operation via `sqlx` or `diesel-async` or an HTTP request).

If this doesn't fit your use-case, you have to drop down to the mid-level API, as explained above.

lib/blowocking/Cargo.toml
```diff
@@ -1,14 +1,14 @@
...
 version.workspace = true
-license.workspace = true
+license = "MIT OR Apache-2.0"
 
 [dependencies]
 once_cell = "1.19.0"
 rayon = "1.8.1"
-thiserror = "1.0.56"
+thiserror = "1.0.57"
 tokio = { version = "1.36.0", features = ["rt", "sync"] }
```
aumetra, line 12: The `rt` feature should probably be made optional by gating the I/O blocking function behind a feature flag.

For the `http-signatures` crate we only need the `sync` feature

aumetra:
Note: Max age of signatures will be 15min. Writing it down here so I don't forget

lib/http-signatures/src/cavage/safety_check.rs
```diff
@@ -0,0 +1,108 @@
...
+    // Check if the `headers` field either includes `date` or `(created)`
+    if !collected_headers.contains(&"date") && !collected_headers.contains(&"(created)") {
+        return Err(SafetyCheckError::MissingRequiredHeaders);
+    }
+
+    if let Some(created) = signature_header.created {
+        let created_time = SystemTime::UNIX_EPOCH + Duration::from_secs(created);
+        let now = SystemTime::now();
+
+        if now.duration_since(created_time)? > MAX_ACCEPTED_SIGNATURE_AGE {
```
perillamint, line 73: What if we have explicit `(expires)` pseudo header? should we follow it or should we `min(MAX_ACCEPTED_SIGNATURE_AGE, expires)`?

lib/http-signatures/src/cavage/safety_check.rs
```diff
@@ -0,0 +1,108 @@
...
+            return Err(SafetyCheckError::SignatureTooOld);
+        }
+    }
+
+    if let Some(date_header) = req.headers().get(DATE) {
+        let date_header_time = httpdate::parse_http_date(date_header.to_str()?)?;
+        let now = SystemTime::now();
+
+        if now.duration_since(date_header_time)? > MAX_ACCEPTED_SIGNATURE_AGE {
```
perillamint, line 82: Potential edge case: `(created)` + `Date`. Which should we trust? 

Also, before checking it, I think it should be checked signature includes it or not.

perillamint (commented):
First batch of review. I'll keep looking the code

lib/http-signatures/src/cavage/safety_check.rs
aumetra: I'd go for a `min` in that case The smaller the time window, the better

lib/http-signatures/src/cavage/easy.rs
```diff
@@ -0,0 +1,124 @@
...
+{
+    // First, set/overwrite the `Date` header
+    let date_header_value =
+        HeaderValue::from_str(&httpdate::fmt_http_date(SystemTime::now())).unwrap();
+    req.headers_mut().insert(DATE, date_header_value);
+
+    let headers = match *req.method() {
+        Method::GET => GET_HEADERS.iter().copied(),
+        Method::POST => POST_HEADERS.iter().copied(),
+        _ => return Err(Error::UnsupportedHttpMethod),
```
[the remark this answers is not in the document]
aumetra: Yep. We need better documentation anyway.. there is literally none right now

lib/http-signatures/src/cavage/safety_check.rs
perillamint: I think... we can use `min((created), Date)` in here, This should be counted on the calculation of `(expires) - (signature created timestamp)`.

Also, it should be checked against time-traveling scenario. (We should not accept request from the future)

perillamint:
Also, general comment. How long should we allow a "future" timestamp? We are not running a fleet of the whole fediverse in a perfectly synchronized data center... I think we should accept... one or two seconds of hiccups.

lib/http-signatures/src/cavage/safety_check.rs
aumetra: Added a check

lib/http-signatures/src/cavage/safety_check.rs
aumetra: Even though.. worst case you could only get through one of the checks without getting your signature rejected.  
Let's say you have a `Date` header which is correct and a `(created)` field that is attacker controlled.

Then the first check, the `(created)` check, might work through the attacker adding an invalid timestamp, but the signature will be discarded when checking the `Date` header (or vice versa).

So I'd say the current code is logically sound, even if one check is technically redundant.

aumetra: Unless one attack vector one wants to proof against is the malicious modification of incoming activities to break federation for you. But I think that attack vector is a bit out there since it requires a MITM position not only in your data centre's packets but also in your TLS connection with other instances.

aumetra:
Does that look good @perillamint ?

perillamint:
LGTM!

PR #490, opened 2024-02-17
WASM-based MRF

aumetra:
This implementation is based on my WIP FEP and will act as its reference implementation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -102,6 +102,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0942ffc6dcaadf03badf6e6a2d0228460359d5e34b57ccdc720b7382dfbd5ec5"
 
 [[package]]
+name = "ambient-authority"
+version = "0.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ec8ad6edb4840b78c5c3d88de606b22252d552b55f3a4699fbb10fc070ec3049"
+
+[[package]]
 name = "android-tzdata"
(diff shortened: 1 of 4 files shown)


aumetra:
Okay, so I'm probably gonna use WIT instead of handrolling the types. Might give better performance (less allocations) and easier portability via the codegen.  
I'm just interested in how this integrates with async/await.

aumetra:
Note: The `mrf-manifest` (and probably `example-mrf`) should be moved to its own repository once the API has been stabilized.

aumetra:
Okay, most ideas are now encoded into the manifest. I think it's in the state where I can think about updating the draft FEP and moving things to own repositories.

aumetra:
Note: It is intentional that `serde_json` is used instead of `simd-json` since `serde_json` allows operating on immutable byte slices. Another reason is the canonical JSON formatter which is inherently tied to `serde_json`.

Both can be addressed once we switch to `sonic-rs` (see #493), since it:

- provides a `Formatter` infrastructure
- allows to operate on immutable slices

aumetra:
So, we now have a manifest definition along with its JSON Schema and CLI tooling to create MRF modules.  
Now we need the configurability via the configuration file, and that should be it. I hope.

Well, I still need to put stuff into the form of an FEP

aumetra:
Okay, now that we got the logging facade, now on to the promised K/V store.

aumetra:
Still needs a redis backend for the K/V storage

aumetra:
Okay, after a little sidequest to remove `futures-retry-policies`, we are back

PR #505, opened 2024-03-17
Extend test coverage

aumetra:
Extending the test coverage of the following sub-crates:

- `geomjeungja`
- `masto-id-convert`
- `speedy-uuid`
- `tower-http-digest`


aumetra:
Last piece I want to test is the `mrf-manifest` crate and its parsing functionalities.  
Luckily that will be pretty easy using the `wat` crate and some manual WASM in text format.

Similar testing approach to what we use at CosmWasm

aumetra:
TODO: Add standalone `speedy-uuid` serialize test

PR #512, opened 2024-03-27
Nix flake

haraldh:


haraldh:
Want me to disable `cargo test` in the "Nix checks" action? There is a `cargo nextest` already, and you probably don't want to edit ´flake.nix` constantly to disable tests, which need docker

aumetra:
Yeah, I think testing inside the nix build isn't really needed since we have our CI to ensure buildability.

haraldh:
Done as a separate commit, so it is easy to revert.

flake.nix
```diff
@@ -15,96 +15,133 @@
...
+            NIX_OUTPATH_USED_AS_RANDOM_SEED = "aaaaaaaaaa";
           };
+
+          cargoTestExtraArgs = pkgs.lib.strings.concatStringsSep " " [
+            "--"
+            # Depend on creating an HTTP client and that reads from the systems truststore
+            # Because nix is fully isolated, these types of tests fail
+            #
+            # Some (most?) of these also depend on the network? Not good??
+            "--skip=activitypub::fetcher::test::federation_allow"
```
aumetra, line 80: Since we don't run tests anymore, can we remove these arguments?

haraldh: Good point! Done!

aumetra:
Thank you for introducing the crane library, I saw a bit about it but never really looked into it too much.  
About the Nix build though, is it doing a full release build? The timings seem like it. Can we downgrade that to a debug build maybe?

haraldh:
I had a misconfig in the crane calls, so it did not use the cached dependency build... Fixed that. Additionally the last commit changes the default to the `dev` profile, which is the `debug` mode. This although populates the github cache with lots more payload, because the debug artifacts are way bigger.

haraldh:
Don't know why the build has failed.. no logs available. Maybe it ran out of RAM??

aumetra:
Just gonna rerun it, maybe it works now? If not, then I'll have a look.  
Yeah, I'd assume either RAM or disk space. Either way it's not good. I really need to start profiling the compilation and figure out where all the RAM is going

haraldh:
I had to change the nix build TMPDIR for my system, because it builds in `/tmp` and this is a tmpfs

haraldh:
```
kitsune-cli> installing /build/source/target/debug/kitsune-cli
kitsune-cli> cp: error copying '/build/source/target/debug/kitsune-cli' to '/nix/store/yjnadx1arhqmg9yj4y955w8ar3x8hnyr-kitsune-cli-0.0.1-pre.6/bin/kitsune-cli': No space left on device
error (ignored): error: cannot unlink '/tmp/nix-build-kitsune-0.0.1-pre.6.drv-0/source/target/debug/deps': Directory not empty
error: writing to file: No space left on device
```

Seems like a 2 debug builds in parallel is too big :)

aumetra (approved):

aumetra:
Thank you for working on this! This looks really good

PR #517, opened 2024-04-03
Rework error representation

### LINKED ISSUE #500: Revamp internal error representation
It's in a horrible state, quite frankly. There doesn't need to be an enum case for each error, we can have a catch-all and a more structured approach to things where we need more info (i.e. Diesel or garde).

aumetra:
Closes #500 

So, for the beginning I simply collapsed all the enum cases into a single "Other" case which has the representations of `eyre::Report` which is fairly compact from a memory usage standpoint.

That will drop the stack usage of errors down significantly (by at least 4x).

I'm still not convinced whether this is the correct solution.


aumetra:
`kitsune-messaging` is completely underutilized. Either we fix this up soon or we remove it.  
I'd even opt to remove it for now. We don't need it yet. And in its current state, it's not really fit for usage anyway.

aumetra:
Removed `kitsune-messaging` for now. We gotta re-add it once we implement WebSocket stuff for the streaming API.  
But until then, bye bye!

aumetra:
(well, damn. Of course coverage increases if we delete ~1k LoC)

PR #520, opened 2024-04-09
Decompose the job scheduler

### LINKED ISSUE #501: Remove Redis dependence
**Is your feature request related to a problem? Please describe.**
Kitsune right now is dependent on Redis even though it really doesn't need to be. All of its functionality can be replaced by PostgreSQL with maybe 1 or 2 extensions (where `pg-trunks` is an excellent way to install them)

**Describe the solution you'd like**
Add a mode for the job scheduler where it can run without Redis and instead with something like a combination of `pgmq` and `pg_cron`. Not sure about the choice of plugins yet, it's subject to bikeshedding.

aumetra:
This PR decomposes the job scheduler into multiple smaller parts, move it to be trait-based, and support an alternative database-based backend.

TODO:

- [x] Decompose


aumetra:
Oh wow, a single test and we get a project coverage increase by 1.34%?

PR #531, opened 2024-05-09
Setup cargo-deny

aumetra:


aumetra:
Ugh.. almost forgot.. because `utoipa` now tries to dynamically download the file, Nix builds fail because that's everything *but* deterministic..

PR #533, opened 2024-05-11
Lower dependency count

aumetra:
This PR just gets rid of some dependencies across the workspace to slim down the gigantic dependency tree.

Open question:

- [x] Do we want to get rid of OpenTelemetry for now, considering we have no production deployments yet, and reintroduce it later when `tonic` was upgraded to http v1, which is one of the sources of the most duplicate dependencies at the moment (see <https://github.com/hyperium/tonic/pull/1670> and <https://github.com/open-telemetry/opentelemetry-rust/pull/1674>)


aumetra:
We are removing OpenTelemetry for now :(  
Would prefer not to have to do this, but `tonic` just pulls in so many duplicate dependencies..

aumetra:
Actually, I have an idea.. this will be hacky.

aumetra:
Okay, that should be fine. Patched every OpenTelemetry dependency and the tonic dependency.. geez.

PR #535, opened 2024-05-13
Replace `redis-rs` with `fred`

### LINKED ISSUE #534: Replace `redis-rs` with `fred`
It's less complex, has less feature flags, and features integrated pooling (which allows us to get rid of `multiplex-pool`), and has a simple interface for clustering.

Seems like an all-round upgrade. And also `fred` is built only for `tokio`, meaning it features less kinda awkward abstractions that try to support multiple runtimes.

Ref: <https://docs.rs/fred/latest/fred/>

aumetra:
Closes #534


aumetra:
I don't know *how* but somehow we broke everything lol. Investigating why the tests now loop indefinitely.

PR #545, opened 2024-06-06
fix(docs): fix config example for max-upload-size.

zicklag:
The format was changed to parse sizes instead of integer number of bytes.


zicklag:
Oh, there are more missing configs in there. Didn't bother keeping track of them all now.

aumetra:
Yeah, don't worry about that. I'll file an issue so I don't forget and then fix it later

PR #549, opened 2024-06-12
Add tests for the keeper

aumetra:
Since this is kinda vital to the job scheduler and a section where we use unsafe code, I feel like we should test this better.


aumetra:
Tests also pass miri just fine

PR #558, opened 2024-07-20
Dependency update

aumetra:
It's been a while, so here it is: A full dependency update for Kitsune.


aumetra:
Another sideeffect of this update is that it makes the Nix checks in CI pass again which broke because of all the dependency patching with OpenTelemetry.

So that's nice!

PR #601, opened 2024-10-25
New OAuth2 implementation

aumetra:
Move away from `oxide-auth` to a homegrown impl


aumetra:
Okay, two large ToDos still:

- [ ] Add extensive tests
- [ ] Make errors happen standards-compliant

aumetra:
I'm actually not too confident that the solution I came up with for handling errors is the best, to be honest..

Maybe I'll just do it differently (have an internal function that does simple `?` error handling, and then an entry function that does the manual error conversion. Would get rid of the macros and result in maybe cleaner code?

aumetra:
TODO: Port some tests from `oxide-auth` (ofc with attribution and all): <https://github.com/HeroicKatora/oxide-auth/tree/master/oxide-auth/src/endpoint/tests>

aumetra:
🙏 one more file to cover. the `flow/mod.rs` with the dispatch. shouldn't take too long. i wanna get this over with so i can focus on more fun things

PR #607, opened 2024-11-11
Build Docker images directly with Nix

aumetra:
In my relentless quest to use Nix for everything: behold, 100% space efficient Docker images.

This produces a Docker image that is 333MB large. Maybe not perfectly efficient sounding, but considering the dynamically linked libraries the GNU target needs, this is completely reasonable IMO.

Adds the new outputs:
- `cli-docker`
- `job-runner`
- `job-runner-docker`
- `main-docker`

Previously the regular `main` output included the job runner binary. Not anymore.


aumetra:
Using the `dist` profile cut off around 89MB. Probably through slimming down the binary and getting rid of some dynamically linked libraries that weren't actually needed.

Image is now 244MB large. Good enough.

aumetra:
Probably a bit more optimizable if we used musl instead, and tried to get rid of all dynamically linked stuff, and inspected what Nix is actually packaging. But this is good enough because we previously didn't have any guaranteed-to-build Docker images.

PR #643, opened 2025-02-18
Small misc changes and opts

aumetra:


aumetra:
Also, yeah, it now includes a `tokio-console` integration

PR #658, opened 2025-03-18
More frontend things

aumetra:
- Extract hero into component
- Add users page
- Style fixes
- Dismiss drawer on navigation
- Hangaku only on 2XL displays (previously large displays)
- Negative top margin only on large displays


aumetra:
The medium image zoom thing is actually not that hard to implement. Just used a library for now because I'm lazy, replacing with an own impl soon

aumetra:
Pagination is a lil wonky. Gotta work on that, I guess.

PR #669, opened 2025-04-14
UI progress

aumetra:
## Changes

- Add Melt-UI (both old and new, since the new one fixes some bugs of the old one but the new one isn't feature complete yet. It's a mess.)
- Install `svelte-fluent` for i18n
- Fix some subtle bugs (zoom didn't work correctly, visibility dropdown was really bad)


aumetra:
Oh yeah, forgot that it can't do Svelte. Great, so I'm alone on that one :upside_down_face:

aumetra:
And yes, Melt-UI + daisyUI is really neat

